PR #357, opened 2020-07-08
add menu option to optionally show grid coordinates in side view; closes #102

### LINKED ISSUE #102: add menu option to optionally show grid coordinates in side view
Currently, only the helix index is displayed.

If this is enabled, the grid coordinates should be displayed in smaller font below the helix index.

rishabhmudradi:
Adds menu option under view named 'Show grid coordinates in side view' to display grid coordinates in smaller font under helix indexes.

### DIFF
--- a/lib/src/view/design_side_helix.dart
+++ b/lib/src/view/design_side_helix.dart
@@ -62,7 +62,6 @@ class DesignSideHelixComponent extends UiComponent2<DesignSideHelixProps> with P
     // of idx, which is useful for making figures in the documentation showing how the grids work
 //    bool SHOW_HELIX_COORDINATES_INSTEAD_OF_IDX = true;
     bool SHOW_HELIX_COORDINATES_INSTEAD_OF_IDX = false;
-    bool SHOW__INDEX_AND_COORDINATES = props.show_grid_coordinates;
     var temp_text_for_help_doc_figure_making;
 
     String tooltip;
@@ -94,13 +93,13 @@ class DesignSideHelixComponent extends UiComponent2<DesignSideHelixProps> with P
               ? temp_text_for_help_doc_figure_making
               : props.helix.idx.toString(),
           Dom.svgTitle()(tooltip)),
-      (Dom.text()
+      if (props.show_grid_coordinates)
+        (Dom.text()
             ..style = {'font-size': 10}
             ..dominantBaseline = 'text-before-edge'
             ..textAnchor = 'middle'
-            ..y = '10'
-            ..key = 'text')(
-          SHOW__INDEX_AND_COORDINATES ? temp_text_for_help_doc_figure_making : "",),
+            ..y = constants.HELIX_RADIUS_SIDE_PIXELS/2
+            ..key = 'text')(temp_text_for_help_doc_figure_making),
           (Dom.svgTitle()(tooltip)),
     ];
 
--- a/lib/src/view/design_side_helix.dart
+++ b/lib/src/view/design_side_helix.dart
@@ -63,17 +63,18 @@ class DesignSideHelixComponent extends UiComponent2<DesignSideHelixProps> with P
 //    bool SHOW_HELIX_COORDINATES_INSTEAD_OF_IDX = true;
     bool SHOW_HELIX_COORDINATES_INSTEAD_OF_IDX = false;
     var temp_text_for_help_doc_figure_making;
+    String grid_position_str;
 
     String tooltip;
     if (props.grid.is_none()) {
       var pos = props.helix.position3d();
       int precision = constants.NUM_DIGITS_PRECISION_POSITION_DISPLAYED;
       tooltip = '${pos.x.toStringAsFixed(precision)}, ${pos.y.toStringAsFixed(precision)}';
-      temp_text_for_help_doc_figure_making = '${pos.x.toStringAsFixed(1)},${pos.y.toStringAsFixed(1)}';
+      grid_position_str = '${pos.x.toStringAsFixed(1)},${pos.y.toStringAsFixed(1)}';
     } else {
       var pos = props.helix.grid_position;
       tooltip = '${pos.h}, ${pos.v}';
-      temp_text_for_help_doc_figure_making = tooltip.replaceAll(' ', '');
+      grid_position_str = tooltip.replaceAll(' ', '');
     }
 
     var children = [
@@ -90,7 +91,7 @@ class DesignSideHelixComponent extends UiComponent2<DesignSideHelixProps> with P
             ..onClick = ((e) => this._handle_click(e, helix))
             ..key = 'text')(
           SHOW_HELIX_COORDINATES_INSTEAD_OF_IDX
-              ? temp_text_for_help_doc_figure_making
+              ? grid_position_str
               : props.helix.idx.toString(),
           Dom.svgTitle()(tooltip)),
       if (props.show_grid_coordinates)
@@ -99,7 +100,7 @@ class DesignSideHelixComponent extends UiComponent2<DesignSideHelixProps> with P
             ..dominantBaseline = 'text-before-edge'
             ..textAnchor = 'middle'
             ..y = constants.HELIX_RADIUS_SIDE_PIXELS/2
-            ..key = 'text')(temp_text_for_help_doc_figure_making),
+            ..key = 'text')(grid_position_str),
           (Dom.svgTitle()(tooltip)),
     ];
 
--- a/lib/src/view/design_side_helix.dart
+++ b/lib/src/view/design_side_helix.dart
@@ -33,6 +33,7 @@ mixin DesignSideHelixProps on UiProps {
   bool selected;
   bool mouse_is_over;
   bool helix_change_apply_to_all;
+  bool show_grid_coordinates;
   MouseoverData mouseover_data;
   BuiltSet<EditModeChoice> edit_modes;
   Grid grid;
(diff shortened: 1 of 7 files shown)


lib/src/view/design_side_helix.dart
```diff
@@ -61,6 +62,7 @@ class DesignSideHelixComponent extends UiComponent2<DesignSideHelixProps> with P
     // of idx, which is useful for making figures in the documentation showing how the grids work
 //    bool SHOW_HELIX_COORDINATES_INSTEAD_OF_IDX = true;
     bool SHOW_HELIX_COORDINATES_INSTEAD_OF_IDX = false;
+    bool SHOW__INDEX_AND_COORDINATES = props.show_grid_coordinates;
```
dave-doty, line 65: Don't make this intermediate variable `SHOW__INDEX_AND_COORDINATES`.

The convention is that all-caps things like this are constants that don't change while the program is running. Some of them are constants used in normal running of the program (in lib/src/constants.dart) and others are used for us to temporarily test something (such as `SHOW_HELIX_COORDINATES_INSTEAD_OF_IDX`, which I used to make some figures, or `USE_REDUX_DEV_TOOLS` and `RUN_TEST_CODE_INSTEAD_OF_APP` at the top of app.dart). Just use `props.show_grid_coordinates` directly wherever it is needed.

You'll find some code that assigns props variables into local variables, e.g., `var helix = props.helix;`, but in general I prefer to reference `props` variables directly (i.e., just use `props.helix`), which helps to see where the data is coming from. This is old code that I used to adapt from an earlier version (prior to using React) where there were already local variables, and it was easier simply to initialize them differently, than to get rid of them and replace them.

lib/src/view/design_side_helix.dart
```diff
@@ -92,6 +94,14 @@ class DesignSideHelixComponent extends UiComponent2<DesignSideHelixProps> with P
               ? temp_text_for_help_doc_figure_making
               : props.helix.idx.toString(),
           Dom.svgTitle()(tooltip)),
+      (Dom.text()
```
dave-doty, line 97: Instead of always rendering the text element, and giving it an empty string sometimes, it's cleaner simply not to render it at all if `props.show_grid_coordinates` is false. Replace this with 

```dart
if (props.show_grid_coordinates)
  (Dom.text() 
  // ...
```

That's the [collection if](https://dart.dev/guides/language/language-tour#collection-operators) operator. (explained more here: https://medium.com/dartlang/announcing-dart-2-3-optimized-for-building-user-interfaces-e84919ca1dff)

lib/src/view/design_side_helix.dart
```diff
@@ -92,6 +94,14 @@ class DesignSideHelixComponent extends UiComponent2<DesignSideHelixProps> with P
...
+            ..style = {'font-size': 10}
+            ..dominantBaseline = 'text-before-edge'
+            ..textAnchor = 'middle'
+            ..y = '10'
```
dave-doty, line 101: Instead of hard-coding the value 10 here, I'd specify it as a fraction of `constants.HELIX_RADIUS_SIDE_PIXELS`. That way if we change that constant later, the text will still go to the same relative position of the helix. (Of course, the font size is hard-coded and will look bigger or smaller relative to the helix, but we can deal with that in a separate issue where we let the user control the font size themselves.)

lib/src/view/menu.dart
```diff
@@ -55,7 +55,8 @@ UiFactory<MenuProps> ConnectedMenu = connect<AppState, MenuProps>(
     ..display_major_tick_widths_all_helices = state.ui_state.display_major_tick_widths_all_helices
     ..invert_y_axis = state.ui_state.invert_y_axis
     ..show_helix_circles_main_view = state.ui_state.show_helix_circles_main_view
-    ..warn_on_exit_if_unsaved = state.ui_state.warn_on_exit_if_unsaved),
+    ..warn_on_exit_if_unsaved = state.ui_state.warn_on_exit_if_unsaved)
```
dave-doty, line 58: This looks wrong. The parentheses should be after the last props variable is set. Now that `show_grid_coordinates_side_view` is the last, the last two lines should be 

```dart
    ..warn_on_exit_if_unsaved = state.ui_state.warn_on_exit_if_unsaved
    ..show_grid_coordinates_side_view = state.ui_state.show_grid_coordinates_side_view),
```

I'm surprised this even compiles. You've tested this and it works?

dave-doty:
One more thing I forgot to mention: the default value of `show_grid_coordinates_side_view` in app_ui_state.dart should be false.

lib/src/view/menu.dart
rishabhmudradi: Not sure, when I ran both the unit tests and the live server everything looked like it was working. I'll change this and the rest of the issues and re-update.

dave-doty: Yes, I checked out this commit and was able to run it. I think it works because the parentheses surround the `Menu()` constructor and subsequent field assignments, and the value of the expression in the parentheses is `Menu`, so it's possible to assign to its fields outside of the parentheses. I think maybe the parentheses are unnecessary in this case.

But outside of a `connect` function, the typical way to make a React component looks like this:

```dart
return (Dom.rect()
          ..x = background_x
          ..y = background_y
          ..width = background_width
          ..height = background_height
          ..className = 'insertion-background'
          ..onClick = ((_) => change_insertion_length())
          ..key = 'rect')();
```

If I didn't set any of those properties, and removed the parentheses around `Dom.rect()`, it would look like this:

```dart
return Dom.rect()();
```

`Dom.rect()` returns an object of type `UiProps` (in this case `SvgProps`, but for custom components, it's the type of the `Props` mixin, e.g., for `MenuBoolean` it's of type `MenuBooleanProps`, whose fields are listed in `MenuBooleanPropsMixin`). That's the object that you set the properties on. Once you're done, that object is callable like a function, so you call it by putting parentheses again. Either it takes no argument, or the arguments specified the children in the View tree (if you look around, you can see some `Dom.g` and `Dom.div` elements always have arguments in the second set of parentheses). The return value of *that* call is the actual `ReactElement` object.

If you tried without the parentheses when specifying one or more props, then it wouldn't work:

```dart
//XXX: compilation error, don't do this
return Dom.rect()
          ..x = background_x
          ..y = background_y
          ..width = background_width
          ..height = background_height
          ..className = 'insertion-background'
          ..onClick = ((_) => change_insertion_length())
          ..key = 'rect'();
```

I think the difference is that in `connect` and `mapStateToProps`, you don't do the second set of parentheses, because `mapStateToProps` is a function that returns a `UiProps` object, not a `ReactElement`. But I just got in the habit of putting parentheses around the whole thing anyway.

lib/src/view/design_side_helix.dart
```diff
@@ -62,7 +62,6 @@ class DesignSideHelixComponent extends UiComponent2<DesignSideHelixProps> with P
     // of idx, which is useful for making figures in the documentation showing how the grids work
 //    bool SHOW_HELIX_COORDINATES_INSTEAD_OF_IDX = true;
     bool SHOW_HELIX_COORDINATES_INSTEAD_OF_IDX = false;
-    bool SHOW__INDEX_AND_COORDINATES = props.show_grid_coordinates;
     var temp_text_for_help_doc_figure_making;
```
dave-doty: It's fine to use this variable, since it already is set based on the same logic I used for making the figure. But change its name to something more appropriate to its current function. I'd just suggest just `grid_position_str` (i.e., the helix's grid position, as a string).

Also I like to put the explicit type declaration when declaring the variable, if it is not assigned right away, i.e., `String grid_position_str;`. That helps to see its type since it has no initial value to help us see it.

dave-doty:
Here's something I overlooked. There's a React duplicate key error, if you check the browser DevTools (Ctrl+Shift+I on Chrome):

image

Here's the source of it, lines 92 and 103, where two text elements both have the key `'text'`:

https://github.com/UC-Davis-molecular-computing/scadnano/blob/d73ee66eb2ba5967a067960a3220df560ecb564f/lib/src/view/design_side_helix.dart#L87-L103

Make sure to check the DevTools console to see if there are any errors. Sometimes they aren't apparent just from looking at the main part of the browser.

dave-doty:
There was also this error when grid coordinates are shown:

image

The cause is this line:

https://github.com/UC-Davis-molecular-computing/scadnano/blob/d73ee66eb2ba5967a067960a3220df560ecb564f/lib/src/view/design_side_helix.dart#L99

There's a different syntax in Dart for assigning this sort of stuff than in Javascript. I'm not sure why the line above that I wrote worked; perhaps it worked on a previous version, back when I used that hack to make the figure.

Here's what that line should be instead:

```dart
..fontSize = 10
```

I can make it work by switching 

```dart
..style = {'font-size': 10}
```

to 

```dart
..style = {'fontSize': 10}
```

But it's better to do

```dart
..fontSize = 10
```

because it's more strongly typed. If you spell the name `fontSize` incorrectly, the compiler warns you. Using the string `'font-size'`, the error doesn't happen until we run the code.

dave-doty:
Ah, another thing I missed. Since `show_grid_coordinates_side_view` is a setting we want to store in localStorage, in order for it to be automatically written, you have to not only put it in `state.ui_state.storables` (which you did), but also the action(s) changing it have to implement `AppUIStateStorableAction`. So instead of

```dart
abstract class ShowGridCoordinatesSideViewSet
    with BuiltJsonSerializable
    implements Action, Built<ShowGridCoordinatesSideViewSet, ShowGridCoordinatesSideViewSetBuilder> {
```

we have

```dart
abstract class ShowGridCoordinatesSideViewSet
    with BuiltJsonSerializable
    implements AppUIStateStorableAction, Built<ShowGridCoordinatesSideViewSet, ShowGridCoordinatesSideViewSetBuilder> {
```

This makes sure that the action triggers this to happen, which is the middleware responsible for storing everything in localStorage:

https://github.com/UC-Davis-molecular-computing/scadnano/blob/d73ee66eb2ba5967a067960a3220df560ecb564f/lib/src/middleware/local_storage.dart#L111-L113

rishabhmudradi:
Hi Professor should I republish the 'add-menu-option-to-optionally-show-grid-coordinates-in-side-view' branch and create a pull request to dev after I change all these issues?

dave-doty:
No, I fixed them already in the dev branch directly. I just wanted to point it out to help you learn some of the details and gotchas. :)

PR #365, opened 2020-07-10
add option to save/not save DNADesign in localStorage on every edit; closes 348

rishabhmudradi:
Adding feature under 'Files' in menu that allows users to enable or disable saving DNA designs into local storage.

### DIFF
--- a/lib/src/actions/actions.dart
+++ b/lib/src/actions/actions.dart
@@ -2069,17 +2069,17 @@ abstract class ShowGridCoordinatesSideViewSet
 /////////////////////////////////////////////////////////////////////////////////////////////////////////////
 // add option to not save DNADesign in localStorage on every edit
 
-abstract class SaveDNAInLocalStorageSet
+abstract class SaveDNADesignInLocalStorageSet
     with BuiltJsonSerializable
-    implements AppUIStateStorableAction, Built<SaveDNAInLocalStorageSet, SaveDNAInLocalStorageSetBuilder> {
-  bool get save_dna_in_local_storage;
+    implements AppUIStateStorableAction, Built<SaveDNADesignInLocalStorageSet, SaveDNADesignInLocalStorageSetBuilder> {
+  bool get save_dna_design_in_local_storage;
 
   /************************ begin BuiltValue boilerplate ************************/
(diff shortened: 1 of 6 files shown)


lib/src/actions/actions.dart
```diff
@@ -2066,6 +2066,21 @@ abstract class ShowGridCoordinatesSideViewSet
       _$showGridCoordinatesSideViewSetSerializer;
 }
 
+/////////////////////////////////////////////////////////////////////////////////////////////////////////////
+// add option to not save DNADesign in localStorage on every edit
+
+abstract class SaveDNAInLocalStorageSet
```
dave-doty, line 2072: Can we replace "DNA" with "DNADesign"? In other words change the Action to `SaveDNADesignInLocalStorageSet`, and the same for the field wherever it appears.

Just calling it "DNA" makes it sound like it's only the DNA sequence being saved, not the whole design.

lib/src/middleware/local_storage.dart
```diff
@@ -42,8 +41,8 @@ save(AppState state, Storable storable) {
 
   if (value_string != null)
     window.localStorage[storable_key] = value_string;
-  else
-    window.localStorage.remove(storable_key);
+  //else
```
dave-doty, line 44: These can just be removed (instead of commented out).

lib/src/view/menu.dart
```diff
@@ -180,6 +182,16 @@ really want to exit without saving.'''
       (MenuDropdownItem()
         ..on_click = ((_) => props.dispatch(actions.ExportCodenanoFile()))
         ..display = 'Export codenano')(),
+      DropdownDivider({}),
+      (MenuBoolean()
+        ..value = props.save_dna_in_local_storage
+        ..display = 'Save DNA Design in localStorage'
+        ..tooltip = '''\
+Saves DNADesign in localStorage on every edit. Disabling this minimizes the time needed to render large strands.'''
```
dave-doty, line 190: In the tooltip and display, just call it "design", not "DNADesign". The phrase "DNADesign" might not be familiar to the user unless they use the Python scripting interface.

Also, change the phrase "large strands" to "large designs". (A design can be large because it has many small strands.)

PR #385, opened 2020-07-16
make default max_offset on every helix equal to the max end of domain on any helix; closes #351

### LINKED ISSUE #351: make default max_offset on every helix equal to the max end of domain on any helix
Currently, if `Helix.max_offset` is not specified in the JSON, it defaults to the max `Domain.end` of any Domain on that Helix. This leads to ugly helices like this:

image

Instead, make max_offset the same for every Helix, like this:

image

rishabhmudradi:
If the max_offsets of helices are not specified, make the max_offset the same for every helix (by setting it equal to the max end of domain).

### DIFF
--- a/lib/src/state/dna_design.dart
+++ b/lib/src/state/dna_design.dart
@@ -552,24 +552,8 @@ abstract class DNADesign with UnusedFields implements Built<DNADesign, DNADesign
   }
 
   bool has_nondefault_max_offset(Helix helix) {
-    var ends = domains_on_helix(helix.idx).map((ss) => ss.end);
-    int max_end = 0;
-    int greatest_end_val = 0;
-    for (var strand in strands){
-          for(var domain in strand.domains()){
-            if((domain.end > greatest_end_val)){
-              greatest_end_val = domain.end;
-            }
-          }  
-        }
-    if(ends.isEmpty){
-      max_end = constants.default_max_offset;
-    }else if(helix.max_offset == greatest_end_val){
-      max_end = greatest_end_val;
-    }else{
-      max_end = ends.reduce(max);
-    }
-    return helix.max_offset != max_end;
+    int default_max_end = calculate_default_max_offset(strands);
+    return helix.max_offset != default_max_end;
   }
 
   bool has_nondefault_min_offset(Helix helix) {
@@ -1284,25 +1268,7 @@ _set_helices_min_max_offsets(List<HelixBuilder> helix_builders, Iterable<Strand>
     HelixBuilder helix_builder = helix_builders[idx];
 
     if (helix_builder.max_offset == null) {
-      var substrands = helix_idx_to_substrands[helix_builder.idx];
-      var greatest_max_offset = 0;
-      if(substrands.isEmpty){
-        greatest_max_offset = constants.default_max_offset;
-      }
-      else{
-        for (var strand in strands) {
-          for (var domain in strand.domains()){
-            if(domain.end > greatest_max_offset){
-              greatest_max_offset = domain.end;
-            }
-          }
-        }
-      }
-      var max_offset = greatest_max_offset;
-      for (var substrand in substrands) {
-        max_offset = max(max_offset, substrand.end);
-      }
-      helix_builder.max_offset = max_offset;
+      helix_builder.max_offset = calculate_default_max_offset(strands);
     }
 
     if (helix_builder.min_offset == null) {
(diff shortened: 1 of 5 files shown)


rishabhmudradi:
I just went over the unit tests failures and it looks like the tests that were failing had a 'max_offset' already defined even though the only code I wrote was inside the 'if (helix_builder.max_offset == null)' method. When I tried running the unit tests on my own device, they sometimes pass but end up failing the on other occasions. I think I might just need to restart and work backwards with this issue.

dave-doty:
I suspect this might be one of those cases when the tests are fragile, i.e., testing the wrong thing.

Those tests appear to be related to something with insertions, but since they (like a lot of other tests) test for exact equality between two entire AppStates, when we change the definition like this, the test breaks (but it doesn't mean the library logic is bad; it means the test is bad).

If that's correct, then we just need to re-write the tests.

dave-doty:
One other thing I forgot to mention: the JSON has many optional fields, and if they aren't specified, they get some default. This issue is one example.

In those cases, I also want to check when *writing* JSON whether a field has its default value, and if so, then don't bother writing it. This keeps the file easier to read. So in this case, you'll also want to re-write these lines of code in `DNADesign.to_json_serializable`:

https://github.com/UC-Davis-molecular-computing/scadnano/blob/f7364c53a4283a2541903841ce6598117a5f69c0/lib/src/state/dna_design.dart#L554-L570

dave-doty:
Just to be clear, the default `max_offset` is now defined as follows:

- If there are no strands in the design, then its value is `constants.default_max_offset`.

- If there is at least one strand in the system, then its value is the largest value of `Domain.end` for any domain in the system.

lib/src/state/dna_design.dart
```diff
@@ -553,7 +553,14 @@ abstract class DNADesign with UnusedFields implements Built<DNADesign, DNADesign
 
   bool has_nondefault_max_offset(Helix helix) {
     var ends = domains_on_helix(helix.idx).map((ss) => ss.end);
-    int max_end = ends.isEmpty ? constants.default_max_offset : ends.reduce(max);
+    int max_end = 0;
```
dave-doty, line 556: I don't think this works. Previously, the default `max_offset` for a helix depended only on that helix (`constants.default_max_offset` if the helix had no domains, and the maximum end offset of any domain on it otherwise).

Now, the default max_offset is a function of the whole design. This would fail when writing a design like the first one in the issue (#351):

image

It would assume all of the helices have their default value, since they all end just after the last domain, but in fact those helices (other than -9 and -10) do *not* have the default value, so they should have an explicit `max_offset` field written in the JSON. The helices below have the default value:

image

So although the method takes a helix index as input, it cannot reference *only* that helix. (It will be a bit wasteful since those values will be recalculated on every helix, but that's fine.)

rishabhmudradi:
After I went over the test cases, it turns out there was a problem with my code as I did not account for helices with varying min_offsets (hence the failed tested cases). I addressed these issues and tested a variety of designs to ensure that the max_offsets are the same for all helices regardless of their starting positions.

dave-doty:
Great, the tests worked!

lib/src/state/dna_design.dart
```diff
@@ -553,7 +553,22 @@ abstract class DNADesign with UnusedFields implements Built<DNADesign, DNADesign
...
+    int greatest_end_val = 0;
+    for (var strand in strands){
+          for(var domain in strand.domains()){
+            if((domain.end > greatest_end_val)){
+              greatest_end_val = domain.end;
+            }
+          }  
+        }
+    if(ends.isEmpty){
```
dave-doty, line 565: I'm still not sure this is correct.

`ends` is the ends on just this one helix. It could be empty, but if there are ends on *other* helices, then `max_offset` should not be `constants.default_max_offset`. That constant is only used if there are no strands in the whole system.

We'll take this as an opportunity to do [test-driven development](https://en.wikipedia.org/wiki/Test-driven_development).

*Before* fixing the code, *first* write tests to produce this error. If the unit test code makes sense to you, can you first write a test to capture this? In other words, make up a design that has three helices, one with no ends, one with max end offset 10, and another with max end offset 20. The test should verify that all three helices get assigned `max_offset` equal to 20. If the test is written correctly, it will fail when it tests the empty helix in that design.

Then, once the unit test reproduces the error, fix the error in this line of code.

We can meet and I can explain how to do that. I recommend testing only the `max_offset`'s, instead of the style in the current tests, where the entire `AppState` is tested for equality with the expected state.

PR #406, opened 2020-07-21
unselect strands when switching out of select mode; closes #389

### LINKED ISSUE #389: unselect strands when switching out of select mode
If items are selected (e.g., strands) in select mode, and then we switch to pencil mode, the strands are still colored pink as though they were selected.

Update the state to unselect all items when switching out of select mode.

rishabhmudradi:
Updating state to unselect all items when switching out of select mode. Added test case for future reference as well.

### DIFF
--- a/lib/src/view/edit_mode.dart
+++ b/lib/src/view/edit_mode.dart
@@ -32,14 +32,12 @@ class EditModeComponent extends UiComponent2<EditModeProps> with RedrawCounterMi
       ...[for (var choice in EditModeChoice.values) this._button_for_choice(choice)],
     ]);
   }
-  static button_response(EditModeChoice mode){
-    app.dispatch(actions.EditModeToggle(mode));
-    app.dispatch(actions.SelectionsClear());
-  }
-
   ReactElement _button_for_choice(EditModeChoice mode) {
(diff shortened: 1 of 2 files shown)


lib/src/view/edit_mode.dart
```diff
@@ -32,10 +32,14 @@ class EditModeComponent extends UiComponent2<EditModeProps> with RedrawCounterMi
...
   }
+  static button_response(EditModeChoice mode){
+    app.dispatch(actions.EditModeToggle(mode));
+    app.dispatch(actions.SelectionsClear());
+  }
 
   ReactElement _button_for_choice(EditModeChoice mode) {
     return (Dom.button()
-      ..onClick = ((_) => app.dispatch(actions.EditModeToggle(mode)))
+      ..onClick = ((_) => button_response(mode))
```
dave-doty, line 42: Can you put this inline? I prefer that when the event handler is small and only needs to be called once. Since it's more than a one-liner, the format will be 

```dart
..onClick = (_) {
  app.dispatch(actions.EditModeToggle(mode));
  app.dispatch(actions.SelectionsClear());
}
```

PR #424, opened 2020-08-02
Adding StrandMaker class and tests.

rishabhmudradi:
Currently, the tests do not add helices to the design.

### DIFF
--- a/lib/src/state/design.dart
+++ b/lib/src/state/design.dart
@@ -32,9 +32,15 @@ part 'design.g.dart';
 abstract class Design with UnusedFields implements Built<Design, DesignBuilder>, JSONSerializable {
   Design._();
 
-  factory Design({Grid grid = Grid.none}) => Design.from((b) => b
+  factory Design({Map<int, Helix> helices, Grid grid = Grid.none}){
+    if(helices == null){ //if helices are not specified
+      helices = {};
+    }
+    return Design.from((b) => b
     ..groups[constants.default_group_name] =
-        b.groups[constants.default_group_name].rebuild((g) => g..grid = grid));
+        b.groups[constants.default_group_name].rebuild((g) => g..grid = grid)
+    ..helices.replace(helices));
+  }
 
   factory Design.from([void Function(DesignBuilder) updates]) = _$Design;
 
--- a/lib/src/state/design.dart
+++ b/lib/src/state/design.dart
@@ -32,14 +32,17 @@ part 'design.g.dart';
 abstract class Design with UnusedFields implements Built<Design, DesignBuilder>, JSONSerializable {
   Design._();
 
-  factory Design({Map<int, Helix> helices, Grid grid = Grid.none}){
+  factory Design({Iterable<Helix> helices, Grid grid = Grid.none}){
     if(helices == null){ //if helices are not specified
       helices = {};
     }
+    var helices_map = {
+      for (var helix in helices) helix.idx: helix
+    };
     return Design.from((b) => b
     ..groups[constants.default_group_name] =
         b.groups[constants.default_group_name].rebuild((g) => g..grid = grid)
-    ..helices.replace(helices));
+    ..helices.replace(helices_map));
   }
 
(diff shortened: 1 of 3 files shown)


rishabhmudradi:
I noticed that there was already a default constructor for Design that added a grid, should I simply add another parameter for helices there too?

dave-doty:
> 
> 
> I noticed that there was already a default constructor for Design that added a grid, should I simply add another parameter for helices there too?

Yes, that's a good idea.

Make the parameter type `Iterable<Helix>`, that way we can pass in a List, a BuiltList, etc.

rishabhmudradi:
I was getting errors when I intialized the helices paramater as `Iterable<Helix>` so I changed it to be `Map<int, Helix>` type

lib/src/state/design.dart
```diff
@@ -31,9 +32,15 @@ part 'design.g.dart';
 abstract class Design with UnusedFields implements Built<Design, DesignBuilder>, JSONSerializable {
   Design._();
 
-  factory Design({Grid grid = Grid.none}) => Design.from((b) => b
+  factory Design({Map<int, Helix> helices, Grid grid = Grid.none}){
```
dave-doty, line 35: Pass in `Iterable<Helix>` and then convert it to a `Map<int, Helix>`. Use either the `Helix.idx` value to set up the Map, e.g.

```dart
factory Design({Iterable<Helix> helices, Grid grid = Grid.none}){
  var helices_map = {
    for (var helix in helices) helix.idx: helix
  };
  var design = Design.from((b) => b
    ..helices.replace(helices_map)
  );
  // ...
```

Typically I think it will be easier less verbose in unit testing code to pass in a list rather than a map.

rishabhmudradi: Thanks for the help, I just fixed it

test/other_unit_test.dart
```diff
@@ -75,5 +85,200 @@ main() {
 //    expect(design.helices[1].position3d().z, 20);
   });
 
+  group('strand_maker_tests', () {
```
dave-doty, line 88: I'd change this to the following:

```dart
    List<Helix> helices = [
      Helix(idx: 0, geometry: Geometry(), grid_position: GridPosition(0, 0), grid: Grid.square),
      Helix(idx: 1, geometry: Geometry(), grid_position: GridPosition(0, 1), grid: Grid.square),
      Helix(idx: 2, geometry: Geometry(), grid_position: GridPosition(0, 2), grid: Grid.square),
      Helix(idx: 3, geometry: Geometry(), grid_position: GridPosition(0, 3), grid: Grid.square),
      Helix(idx: 4, geometry: Geometry(), grid_position: GridPosition(0, 4), grid: Grid.square),
      Helix(idx: 5, geometry: Geometry(), grid_position: GridPosition(0, 5), grid: Grid.square),
    ];
```

For two reasons:

1. It's common for a method or function to accept an `Iterable` as input, to give users maximum flexibility to provide many different types as input, rather than dictating that it has to be a `List` or it has to be a `Set`. But when making your own local variables, it's more helpful to you for debugging to specify the exact type that it is. It will help the compiler warn you about errors, and allow you to do more. For example, if you wanted to set the 3'rd item, you could do `helices[3] = ...` if it's a `List`, but if it's just an `Iterable`, you can't do that. There's a sort of duality here: when choosing a parameter in a method/function for a user to give to you, make it as general as possible (i.e., it should assume the least about the object that is needed to do the job), but when specifying local variables that you will use (and only you control), make it as specific as possible, to let you do the most with the object.

2. `{ ... }` specifies a `Set`, whereas `[ ... ]` specifies a `List`. https://dart.dev/guides/language/language-tour#sets Although the constructor can iterate over either one of them, with a `List` it's guaranteed to iterate over them in order, whereas `Set`'s are not ordered, so although you listed them in increasing order of `idx`, they may come out in a different order. That doesn't really matter for the constructor, but it may be easier to debug if you know the order in which they will be put it. I also think that Dart `Map`'s maintain keys in the order they were inserted, which would help with debugging many steps down the line.

PR #443, opened 2020-08-08
add view option to show length of loopout next to loopout; closes #426

### LINKED ISSUE #426: add view option to show length of loopout next to loopout
Currently, it is shown in the browser on mouseover. However, this would be useful for making figures that don't show the DNA sequence (which implicitly tells the length of the loopout), to be able to read the fully-specified design from the exported SVG figure.

rishabhmudradi:
Adding an option in the view menu to add loopout lengths in the main view.


lib/src/view/design_main_dna_sequence.dart
```diff
@@ -69,7 +73,14 @@ class DesignMainDNASequenceComponent extends UiComponent2<DesignMainDNASequenceP
         Domain next_dom = this.props.strand.substrands[i + 1];
         if (should_draw_domain(prev_dom, side_selected_helix_idxs, props.only_display_selected_helices) &&
             should_draw_domain(next_dom, side_selected_helix_idxs, props.only_display_selected_helices)) {
-          dna_sequence_elts.add(this._dna_sequence_on_loopout(loopout, prev_dom, next_dom));
+          if(props.show_loopout_length_main_view && (props.show_dna == false)){
```
dave-doty, line 76: make sure that you are using the Dart auto-formatter. (In WebStorm it's Ctrl+Alt+L; in Visual Studio Code I think it might be Shift+Alt+F, but I'm not sure.) I can tell it's not being used because it would put a space between `if` and `(`, for instance, and between `)` and `{` at the end of the line.

It might not seem like a big deal, but I'm quite pedantic about having everything be formatted by the Dart formatter, so that the project looks consistent and will be easier to read by someone who, for example, comes from another Dart project.

(Of course we're using `snake_case` instead of `camelCase`, which breaks [recommended Dart convention](https://dart.dev/guides/language/effective-dart/style#do-name-other-identifiers-using-lowercamelcase), but I port so much code from the Python library, where `snake_case` is the convention, that I thought it's better to keep those the same.)

lib/src/view/menu.dart
```diff
@@ -583,6 +585,15 @@ Shows grid coordinates in the side view under the helix index.'''
         ..onChange = ((_) => props.dispatch(actions.ShowGridCoordinatesSideViewSet(
             show_grid_coordinates_side_view: !props.show_grid_coordinates_side_view)))
         ..key = 'show-grid-coordinates-side-view')(),
+      (MenuBoolean()
+        ..value = props.show_loopout_length_main_view
+        ..display = 'Show loopout length in main view'
+        ..tooltip = '''\
+Loopout length is anchored to loopout when selected in main view.'''
```
dave-doty, line 592: I'd phrase this "When selected, the length of each loopout is displayed next to it."

lib/src/view/menu.dart
```diff
@@ -583,6 +585,15 @@ Shows grid coordinates in the side view under the helix index.'''
         ..onChange = ((_) => props.dispatch(actions.ShowGridCoordinatesSideViewSet(
             show_grid_coordinates_side_view: !props.show_grid_coordinates_side_view)))
         ..key = 'show-grid-coordinates-side-view')(),
+      (MenuBoolean()
+        ..value = props.show_loopout_length_main_view
+        ..display = 'Show loopout length in main view'
```
dave-doty, line 590: Remove "in main view" here and pluralize "lengths".

lib/src/view/design_main_loopout_lengths.dart
```diff
@@ -0,0 +1,70 @@
...
+    }
+  }
+
+  @override
+  render() {
+    if (props.show_loopout_length) {
+      return (Dom.g()..className = 'dna-sequences-main-view')([
+        for (Strand strand in props.strands)
+          if (strand.dna_sequence != null)
+            (DesignMainDNASequence()
```
dave-doty, line 56: This isn't quite what I meant. I would completely separate the components that draw loopout lengths from those that draw DNA sequences. So here, it should be something like

```dart
for (Strand strand in props.strands)
  for (Loopout loopout in strand.loopouts())
    (DesignMainLoopout()
      // ...
```

where `DesignMainLoopout` is its own component that is only responsible for drawing loopout lengths, and there is no code in `DesignMainDNASequence` that has anything to do with drawing loopout lengths.

lib/src/view/design_main_loopout_length.dart
```diff
@@ -0,0 +1,57 @@
...
+  Loopout loopout;
+}
+
+class DesignMainLoopoutProps = UiProps with DesignMainLoopoutPropsMixin;
+
+class DesignMainLoopoutComponent extends UiComponent2<DesignMainLoopoutProps> with PureComponent {
+  @override
+  render() {
+    List<ReactElement> loopout_length_elts = [];
+    loopout_length_elts.add(this._loopout_length(props.loopout));
```
dave-doty, line 35: There's no need to pass `props.loopout` as a parameter to a method in the same class; Just access `props.loopout` from within the method `_loopout_length`.

Sometimes I'll but the code generating the component in a separate *function* outside of the class, if I want to call it from elsewhere. (e.g., the code that creates crossovers and loopouts, since they are called by both normal strands and the view of a strand that is being dragged), and in those cases you might need to pass some props as function parameters.

But if it's a method that will only be called from within this class, don't bother passing props since they can be accessed from within the method itself.

lib/src/view/design_main_loopout_length.dart
```diff
@@ -0,0 +1,57 @@
...
+
+  ReactElement _loopout_length(Loopout loopout) {
+    var start_offset = '50%';
+    var dy = '${0.1 * props.geometry.base_width_svg}';
+
+    Map<String, dynamic> style_map = {'fontSize': '${9}px'};
+    var loopout_length = loopout.loopout_length.toString().split('').reversed.join('');
+
+    SvgProps text_path_props = (Dom.textPath()
+      ..className = 'dna-seq-loopout'
```
dave-doty, line 47: `'dna-seq-loopout'` doesn't seem like a good className for this element. It represents a loopout length, so I'd call it `loopout-length` instead.

Also, I'm now starting to put these strings as constants in constants.dart that begin with `'css_selector_'`:

https://github.com/UC-Davis-molecular-computing/scadnano/blob/dc1e16f3209f71239fff5a672b2b5101f4426927/lib/src/constants.dart#L315-L340

This is especially important for anything that will be referenced in the file [scadnano-styles.css](https://github.com/UC-Davis-molecular-computing/scadnano/blob/master/web/scadnano-styles.css).

In fact, if possible I'd like to figure out a way to have a single place to define these constants that can be referenced from both Dart and from the css file, but I don't know how to do that. But collecting all of the Dart CSS selector strings into one place is helpful.

lib/src/view/design_main_dna_sequence.dart
```diff
@@ -28,8 +28,6 @@ mixin DesignMainDNASequencePropsMixin on UiProps {
   BuiltMap<int, Helix> helices;
   BuiltMap<String, HelixGroup> groups;
   Geometry geometry;
-  bool show_loopout_length;
-  bool show_dna;
 }
 
 class DesignMainDNASequenceProps = UiProps
```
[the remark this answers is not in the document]
rishabhmudradi: Yup, I just copy-pasted the code from the dev branch for both those classes to make sure it was the exact same as before.

lib/src/view/design_main_loopout_length.dart
dave-doty: Note that this is not currently followed uniformly in the source code, but I'm trying to update it whenever I see css selector classes that are not defined by a constant in constants.dart.

rishabhmudradi:
I just noticed that there was a conflict, should I just merge dev into this branch to get rid of it?

dave-doty:
> 
> 
> I just noticed that there was a conflict, should I just merge dev into this branch to get rid of it?

I just did that.

PR #444, opened 2020-08-08
Disable redux dev tools for release

### LINKED ISSUE #399: turn off Redux DevTools automatically when deployed
Currently React DevTools is enabled by a const value in lib/src/app.dart.

Instead, it should be controlled from a source that can be set to false when deployed but true when running locally. This way, we won't have to remember to set that constant to false before committing.

UnHumbleBen:
Fixes #399 (in terms of disabling redux dev tools) react dev tools will still work

### DIFF
--- a/build.yaml
+++ b/build.yaml
@@ -22,6 +22,14 @@
 targets:
   $default:
     builders:
+      build_web_compilers|entrypoint:
+        # These are globs for the entrypoints you want to compile.
+        options:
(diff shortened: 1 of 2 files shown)


UnHumbleBen:
Can I merge this pull request?

dave-doty:
> 
> 
> Can I merge this pull request?

Yes, go ahead.

PR #448, opened 2020-08-10
Fixes #436, Fixes #438

### LINKED ISSUE #436: properly display cached PNG of DNA sequences, when some helix group is translated to negative y position
Now that helix groups can be translated anywhere, the hack in the DNA sequence caching here is no longer working:

https://github.com/UC-Davis-molecular-computing/scadnano/blob/2e6931ad0461b63c1e8fc2ddca53314954a5fd6c/lib/src/view/design_main_dna_sequences.dart#L54-L55

To see how it fails, load the design [origami_cross_tile_strong_strong.zip](https://github.com/UC-Davis-molecular-computing/scadnano/files/5036205/origami_cross_tile_strong_strong.zip). When zoomed out it looks like this:

image

But it should look like this:

image

I think the problem is that helix group "north" has y translation -11.26 nm (i.e., about -339 pixels in SVG coordinates, much bigger than the -50 in the caching hack).

### LINKED ISSUE #438: exported SVG main should set bounding box just large enough to contain elements
This is a bit tricky especially now that we have helix groups that can be translated and rotated.

Also, the DNA sequence rendering seems to depends on the same constants that are causing issue #436.

UnHumbleBen:


lib/src/util.dart
```diff
@@ -1307,18 +1307,20 @@ void svg_to_png_data() {
 
   // Translates dna_sequence_element_copy down so that it's Blob uri captures the topmost dna sequence
   // inside of it's view box.
+  Rect bbox = dna_sequence_element.getBBox();
```
dave-doty, line 1310: Can you double-check that the function `svg_to_png_data()` (which calls `document.getElementById` above) is never being called in either a `render()` method of a React component or in a reducer? Both of those should be pure functions (of their parameters/props) and not directly reference "global" information such as DOM elements returned from `document.getElementById`.

UnHumbleBen: Only place it is called is here
https://github.com/UC-Davis-molecular-computing/scadnano/blob/dc1e16f3209f71239fff5a672b2b5101f4426927/web/index.html#L284

Attached as a callback to the `main_view_svg_pan_zoom` object here:
https://github.com/UC-Davis-molecular-computing/scadnano/blob/dc1e16f3209f71239fff5a672b2b5101f4426927/web/index.html#L328

PR #463, opened 2020-08-25
only selects DNA ends on left-click; closes #458

### LINKED ISSUE #458: do not select objects on right-click
Currently, if DNA ends are selectable, and one right clicks on a DNA end to bring up the context menu, it becomes selected:

image

This is incorrect behavior; only left-click should select.

rishabhmudradi:


lib/src/view/design_main_strand_dna_end.dart
```diff
@@ -159,6 +159,9 @@ class DesignMainDNAEndComponent extends UiComponent2<DesignMainDNAEndProps> with
     if (end_selectable(dna_end)) {
       // select end
       MouseEvent event = event_synthetic.nativeEvent;
+      if(event.button == 2){
```
dave-doty, line 162: Can you replace the literal `2` with `constants.RIGHT_CLICK_BUTTON`? Then it's easier to tell what this line of code is doing.

Also can you test prevent middle-click from selecting as well? Currently we don't use middle-click for anything, but in the future we might, and this would help prevent the behaviors from colliding.

PR #466, opened 2020-08-26
Adding code for pointer_up

rishabhmudradi:


lib/src/view/design_main_strand_dna_end.dart
```diff
@@ -171,6 +171,9 @@ class DesignMainDNAEndComponent extends UiComponent2<DesignMainDNAEndProps> with
   handle_end_pointer_up_select(react.SyntheticPointerEvent event_synthetic) {
     if (end_selectable(dna_end)) {
       MouseEvent event = event_synthetic.nativeEvent;
+      if (event.button == constants.RIGHT_CLICK_BUTTON || event.button == constants.MIDDLE_CLICK_BUTTON) {
```
dave-doty, line 174: Can you put some print statements (or use the debugger) in Firefox and Chrome and see what event.button is when you left, middle, and right click on Mac, and then put that in a comments here? I'll do the same on Windows and Linux and see if they are different and that explains why we had this problem. Then we'll have it written down somewhere how this works.

Do this for both for the mouse/pointer down event and the mouse/pointer up event.

rishabhmudradi:
I just tested this and it looks like on a Mac with chrome, the mouse pointer down method was never accessed when I right-click. On Firefox for example, event.button is printed for both mouse/pointer up and down when I right-click. On the other hand on chrome, the event.button is only printed for mouse/pointer up.

PR #468, opened 2020-08-30
Fixing loopout length text on firefox; closes #462

### LINKED ISSUE #462: make loopout lengths centered on Firefox
On Firefox, loopout lengths are off center:

image

They should look like this (image from Chrome):

image

rishabhmudradi:


rishabhmudradi:
I tested this again and for some reason this code doesn't work for loopout lengths less than 4 which are on the same helix. I'll debug and update this.

lib/src/view/design_main_loopout_length.dart
```diff
@@ -26,8 +27,12 @@ class DesignMainLoopoutComponent extends UiComponent2<DesignMainLoopoutProps> wi
 
   ReactElement _loopout_length() {
     var start_offset = '50%';
-    var dy = '${0.1 * props.geometry.base_width_svg}';
+    var dx = '${0}';
```
dave-doty, line 30: It's okay to do this once in a while (hard-coding constants), but first I want to check whether there's some more elegant way of doing it. Try searching Google for things like "Firefox textPath alignment" or something like that. If there's a more "discrete" way to handle this (such as setting some style property related to text alignment), that's superior to plugging in some hard-coded x offset.

Think of it as the difference between using dominant-baseline: https://developer.mozilla.org/en-US/docs/Web/SVG/Attribute/dominant-baseline#Result compared to achieving the same thing with adding or substracting some number of pixels in the y direction. The former will work no matter what font size you have, or if you re-scale everything by a certain amount, but hard-coded offsets are more fragile and likely to look bad if something changes.

That said, maybe we can't figure it out. There's a few places in the code where I have some ugly hard-coded numbers, e.g., 

https://github.com/UC-Davis-molecular-computing/scadnano/blob/67d59d8703067d5add104f1d1f1547abc256c0d4/lib/src/view/design_main_dna_sequence.dart#L205-L212

But it's not ideal if we can avoid it.

A few other things:

1. Remember to use the auto-formatter to format the code.

2. I noticed that the name of the React component is `DesignMainLoopout`, which is confusing. The filename is design_main_loopout_length.dart, and the convention we are using is to use the same name (without underscores) for the class. Also, it's not a loopout being shown (that's defined by a `DesignMainStrandLoopout` component), just a length of a loopout. Can you rename the component to `DesignMainLoopoutLength`? (and the related classes such the Props)

rishabhmudradi:
> I tested this code again and for some reason it does not format loopout lengths less than 4 which are on the same helix the right way. I'll debug and update this issue.

It looks like the dx attribute makes the text curve along the loopout for the above case. Not sure how to fix this.

Screen Shot 2020-09-05 at 12 04 47 AM

lib/src/view/design_main_loopout_length.dart
dave-doty: Are you sure there's not a way to play with the CSS properties instead of hard-coding a dx value?

Does this work, for instance, if the loopout length is 4 digits instead of 1?

It seems to be as though the problem is that the text is not centered on Firefox. Perhaps the CSS property that makes it centered on Chrome is interpreted differently by Firefox. I'd check to see if there's a way to center the text.

dave-doty (changes requested):

lib/src/view/design_main_loopout_length.dart
rishabhmudradi: I think the problem is that dominant-baseline is interpreted differently in Firefox than on Chrome. I looked online and a lot of responses to this was to use the translate function, but I wasn't able to find fully different way to center text.

dave-doty (changes requested):
This still feels wrong. The height of a base shouldn't influence how this works. I suspect there is a CSS alignment property that should fix this. (for instance, something like [text-align](https://www.w3schools.com/cssref/pr_text_text-align.ASP) or [dominant-baseline](https://developer.mozilla.org/en-US/docs/Web/SVG/Attribute/dominant-baseline)).

PR #480, opened 2020-09-22
Fixes #20, add shadow filter for domains

### LINKED ISSUE #20: elements far from origin disappear on mouseover when selectable
This appears to be related to the SVG filter used to create a shadow around the element.

This only appears on large designs, for instance examples/output-designs/6_helix_bundle_honeycomb.dna when putting the cursor over strands on the far right side.

UnHumbleBen:


lib/src/middleware/edit_select_mode_change.dart
```diff
@@ -17,6 +17,7 @@ const selectable_css_style_non_domain_or_end = {
 };
 
 const selectable_css_style_domain = {
+  'filter': 'url("#shadow")',
```
dave-doty, line 20: Is it possible to do this without turning the domain pink and making it thicker? I was only doing those as a stopgap measure since the shadow filter didn't work.

I tried removing `'stroke': 'hotpink'` and `'stroke-width': '5pt'` (i.e., only add the shadow filter), and for me, the problem came back.

dave-doty (changes requested):

lib/src/middleware/edit_select_mode_change.dart
UnHumbleBen: scadnano-domain-shadow-filter-remove-pink-hover
Strange, seems okay for me?

PR #486, opened 2020-10-12
Fixes #473; side cursor title use zy instead of xy

### LINKED ISSUE #473: Fix side-view cursor title to use y,z coordinate instead of x,y
Set grid to none. When you hover the cursor on the side-view, the cursor title uses (x,y) coordinate. You can tell because the first coordinate is always 0, no matter where the cursor moves. Instead, the cursor title should use (y, z) or (z, y) coordinates

UnHumbleBen:


dave-doty (changes requested):
I think (z,y) makes more sense as an ordering than (y,z), since y is still vertical in the side view.

dave-doty (approved):

PR #487, opened 2020-10-14
Performance improvement for strand selection

UnHumbleBen:
Remove passing in app's "currently moving" state via props and instead
directly access the global model from the view.

This makes it so that only the strand that gets selected is rerendered.


dave-doty (approved):
I haven't tested this myself, so I'm assuming it's been tested.

UnHumbleBen:
Yuip it has been tested. You can check it once it has been deployed to dev. Merging now

PR #507, opened 2020-11-09
0.13.0 

UnHumbleBen:


dave-doty:
Let's wait a bit on this until we've stress-tested the dev branch a bit more.

dave-doty:
Let's implement one more feature before merging this, which is to add links to the old versions under the help menu (just like there is currently a link to the dev version).

UnHumbleBen:
> Let's implement one more feature before merging this, which is to add links to the old versions under the help menu (just like there is currently a link to the dev version).

Do you want this to be automated as well? Or should these links be added manually?

dave-doty:
> > Let's implement one more feature before merging this, which is to add links to the old versions under the help menu (just like there is currently a link to the dev version).
> 
> Do you want this to be automated as well? Or should these links be added manually?

Can you figure out a way to automate it? I hadn't thought about that, but it would be nice if it could be read from some file that is automatically generated.

UnHumbleBen:
A simple idea I have would be in `constants.dart`, we have a Dart list of all the versions we want to provide links for, named something like `scadnano_versions`. In the React Menu component, we can write the logic for iterating through `scadnano_versions` and creating a link for each one.

If we implement it this way, we will need to remember to update both the `CURRENT_VERSION`  and `scadnano_versions`  variable whenever we update the version, which I feel like is a reasonable level of automation. So currently, we would probably have a `scadnano_versions` variable that looks like:
```dart
final scadnano_versions = [
  "0.13.0",
  "0.12.1"
];
```

What do you think of this idea @dave-doty ?

dave-doty:
> A simple idea I have would be in `constants.dart`, we have a Dart list of all the versions we want to provide links for, named something like `scadnano_versions`. In the React Menu component, we can write the logic for iterating through `scadnano_versions` and creating a link for each one.
> 
> If we implement it this way, we will need to remember to update both the `CURRENT_VERSION` and `scadnano_versions` variable whenever we update the version, which I feel like is a reasonable level of automation. So currently, we would probably have a `scadnano_versions` variable that looks like:
> 
> ```dart
> final scadnano_versions = [
>   "0.13.0",
>   "0.12.1"
> ];
> ```
> 
> What do you think of this idea @dave-doty ?

Sounds good. Ideally we'll also have something that can warn us if the latest version is not the highest version on that list, to remind us to update the list.

dave-doty (approved):

dave-doty:
Things to mention in release notes (add link to issue where there is an issue for it; check commits to see when issues are referenced):

- swapped definition of x/z coordinates; should not see any change if only using web interface, but breaking change for Python scripting library: Python scripts using no grid and specifying 3D coordinates manually will have to be re-written
- improved batch editing; can select many edit/delete at once:
  - loopouts/crossovers (could delete before, now can convert between them, or change loopout lengths, in batch)
  - deletions/insertions (deleting in batch, editing insertion length in batch, and adding many deletions/insertions to many helices by pressing Ctrl when adding to one Helix)
  - modifications
  - 5'/3' ends and add same modification to all of them at once
- rope select mode: can draw an arbitrary non-self-intersecting polygon to select many items at once
- old versions hosted at scadnano.ord/ subaddresses; might be useful in case an update breaks some functionality
- anything else major??

PR #508, opened 2020-11-12
Fixes #490; store old versions of scadnano at version-specific URLs

### LINKED ISSUE #490: store old versions of scadnano at version-specific URLs
With a program that someone downloads, if an upgrade breaks something they were doing, they can always download the old version and install that instead.

It would be useful for us to deploy older versions of scadnano for this purpose.

Not *all* older versions, but for instance, we could have an archived version whenever a new minor version is deployed from now on.

http://scadnano.org/v0.12.3
http://scadnano.org/v0.13.5
http://scadnano.org/v0.14.2
...

So that, for instance, if upgrading from 0.13.5 to 0.14.0 breaks for someone, they have an option to keep using 0.13.5.

Since we are about to implement a breaking change in #488, it would be good to implement this before we commit that version to master. I'm not sure how automatic we can make it, but it would be nice if it were completely automatic with Github Pages.

UnHumbleBen:
Tested in Fork: https://unhumbleben.github.io/scadnano/v0.11.3/
Here is example of when version number cannot be parsed: https://github.com/UnHumbleBen/scadnano/actions/runs/358816601
Root deployment still runs, but version url gets skipped and a warning message is fired so that we know we need to fix the version string.

### DIFF
--- a/.github/workflows/gh-pages.yml
+++ b/.github/workflows/gh-pages.yml
@@ -40,6 +40,15 @@ jobs:
       - name: Build into gh-pages repo
         run: PATH="$PATH:/usr/lib/dart/bin" pub global run webdev build -o web:gh-pages-repo
       
+      - name: Retrieve version
+        run: export VERSION=$(cat lib/src/constants.dart | grep  "const String CURRENT_VERSION = .*" | grep -oP "\d+\.\d+\.\d+")
+      - name: Build into gh-pages-repo/VERSION
+        run: |
+         if [ "$VERSION" != "" ]; then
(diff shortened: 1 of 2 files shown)


dave-doty (approved):

dave-doty:
> Tested in Fork: https://unhumbleben.github.io/scadnano/v0.11.3/
> Here is example of when version number cannot be parsed: https://github.com/UnHumbleBen/scadnano/actions/runs/35881660
> Root deployment still runs, but version url gets skipped and a warning message is fired so that we know we need to fix the version string.

I see a 404 error on the second link.

UnHumbleBen:
Fix linked in my comment: https://github.com/UnHumbleBen/scadnano/actions/runs/358816601

PR #510, opened 2020-11-17
Add links to older version

UnHumbleBen:
Implements https://github.com/UC-Davis-molecular-computing/scadnano/pull/507#issuecomment-727685759

### DIFF
--- a/lib/src/constants.dart
+++ b/lib/src/constants.dart
@@ -5,10 +5,14 @@ import 'package:platform_detect/platform_detect.dart';
 
 import 'state/grid.dart';
 
-// WARNING: Do not modify line below, except for the version string.
+// WARNING: Do not modify line below, except for the version string
+//          (and also add new version string to scadnano_versions_to_link).
 const String CURRENT_VERSION = "0.13.0";
 const String INITIAL_VERSION = "0.1.0";
-final scadnano_versions = [
+
+// scadnano versions that we deploy so that older versions can be used.
+// CURRENT_VERSION should be the highest verstion in this list.
+final scadnano_versions_to_link = [
   "0.13.0",
   "0.12.1"
 ];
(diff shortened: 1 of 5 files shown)


UnHumbleBen:
Before merging I want to test this out by deploying v0.12.1 manually

lib/src/constants.dart
```diff
@@ -8,6 +8,10 @@ import 'state/grid.dart';
 // WARNING: Do not modify line below, except for the version string.
 const String CURRENT_VERSION = "0.13.0";
 const String INITIAL_VERSION = "0.1.0";
+final scadnano_versions = [
```
dave-doty, line 11: Let's rename this variable to `scadnano_versions_to_link`, since it's not actually all the scadnano versions.

lib/src/view/menu.dart
```diff
@@ -673,6 +673,16 @@ When selected, the length of each loopout is displayed next to it.'''
 // help menu
 
   help_menu() {
+    List<dynamic> older_versions_links = [for (var version in constants.scadnano_versions) DropdownItem(
```
dave-doty, line 676: Does the React menu component let you make sub menus? Like this:

image


That would be ideal here, to have a submenu named "Other versions", and then put dev and the older versions under it.

We may also want to start doing that with some of the clutter on other menus, for example the four ways to save to localStorage under File might be cleaner to have as a sub-menu, or grouping the various "Show/Hide" options in View under a single "Show/Hide" submenu.

dave-doty (changes requested):

lib/src/view/menu.dart
UnHumbleBen: Resolved in https://github.com/UC-Davis-molecular-computing/scadnano/pull/510/commits/d19ac32b88656de64e26a42c8e96e27cd0ef15ee

Created issue #515 to address the cleanup issue

lib/src/constants.dart
UnHumbleBen: Resolved in https://github.com/UC-Davis-molecular-computing/scadnano/pull/510/commits/d19ac32b88656de64e26a42c8e96e27cd0ef15ee

dave-doty (approved):

PR #516, opened 2020-11-18
Fixes #512; fix strand color picker

### LINKED ISSUE #512: Cannot change staple color
Abstract: can't change staple color on  https://scadnano.org/dev/

How to reproduce:

1. Select any origami example on  https://scadnano.org/dev/
2. Right Click on a staple then "set color" then change the color then "OK"
3. The color doesnt change

I am using Chrome 1.0.4044.138

UnHumbleBen:
JS bindings were not setup properly so function name was not
recognized by dart2js compiled code

### DIFF
--- a/lib/src/view/strand_color_picker.dart
+++ b/lib/src/view/strand_color_picker.dart
@@ -40,7 +40,7 @@ mixin StrandColorPickerState on UiState {
 
 class StrandColorPickerComponent
     extends UiStatefulComponent2<StrandColorPickerProps, StrandColorPickerState> {
-  void handleOnChangeComplete(JSColor color, event) {
+  void handleOnChangeComplete(JSColor color, _) {
     state.color = HexColor(color.hex);
   }
 
--- a/lib/src/view/strand_color_picker.dart
+++ b/lib/src/view/strand_color_picker.dart
@@ -1,3 +1,7 @@
+@JS()
+library scadnano;
+
+import 'package:js/js.dart';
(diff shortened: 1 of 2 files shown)


UnHumbleBen:
After we merge to master, we should merge master to dev so that dev will also get this fix, as well as add version 0.12.2 to list of scadnano links

UnHumbleBen:
I will also need to manually deploy v0.12,2 because the Github Actions for automatically deploying new versions have not been pushed to master yet.

lib/src/view/strand_color_picker.dart
```diff
@@ -36,7 +40,7 @@ mixin StrandColorPickerState on UiState {
 
 class StrandColorPickerComponent
     extends UiStatefulComponent2<StrandColorPickerProps, StrandColorPickerState> {
-  void handleOnChangeComplete(color, event) {
+  void handleOnChangeComplete(JSColor color, event) {
```
dave-doty, line 43: If the event is not being used, better to name it `_`. (Also, do you know its type?)

dave-doty (changes requested):

lib/src/view/strand_color_picker.dart
UnHumbleBen: Not sure what type it event is. Here is the API: https://casesandberg.github.io/react-color/#api-onChangeComplete

dave-doty (approved):

PR #538, opened 2020-11-28
Fixes #489, removes AppState equality checks in UT

### LINKED ISSUE #489: make existing unit tests less fragile by not testing whole AppState for equality
In particular, get rid of the function `expect_app_state_equal`, and instead, for each unit test, test only the behavior that test is intended to measure.

For instance, in HelixPositionSet, test only that the helix positions were properly updated. There is no need to also test that SVG positions were updated according to the Position3D --> SVG position algorithm.

UnHumbleBen:


test/reducer_test.dart
```diff
@@ -2128,15 +2097,20 @@ main() {
...
+  test('Testing_DNAEndsMoveStart', () {
     AppState initial_state = app_state_from_design(simple_helix_no_seq_design);
     AppState actual_state = app_state_reducer(
         initial_state, DNAEndsMoveStart(offset: 0, helix: simple_helix_no_seq_design.helices[0]));
     AppState expect_state = initial_state.rebuild((b) => b.ui_state.dna_ends_are_moving = true);
-    expect_app_state_equal(actual_state, expect_state);
+
+    expect(actual_state.ui_state.changed_since_last_save, false);
+    expect(actual_state.ui_state.dna_ends_are_moving, true);
+    expect(actual_state.undo_redo.undo_stack, BuiltList<Design>());
```
dave-doty, line 2108: There's a lot of lines like this where the Undo and Redo stacks are being tested. It's not clear why. In this example, the test is named "Testing_DNAEndsMoveStart", which is describing a different behavior.

In general, each test should only be testing for certain functionality. The fact that the Undo and Redo stacks happen to be empty after this is an implementation detail of the fact that we started with an AppState that happened to have empty Undo and Redo stacks, and DNAEndsMoveStart is not a Design-changing Action.

This is just one example. In general, each unit test should be testing a particular piece of functionality. We might want a separate set of tests for Undo/Redo specifically, but outside of those, we shouldn't be checking the Undo/Redo stacks. Imagine if we change how they are represented. Then all these tests would have to be re-written, even though they aren't testing Undo/Redo behavior.

dave-doty (changes requested):

test/reducer_test.dart
```diff
@@ -2003,17 +1965,13 @@ main() {
...
-      ..ui_state.changed_since_last_save = true);
-    ;
 
     AppState original_state = app_state_from_design(simple_strand_design)
         .rebuild((b) => b..ui_state.storables.side_selected_helix_idxs = SetBuilder<int>([0, 1]));
 
     AppState final_state = app_state_reducer(original_state, HelixRemoveAllSelected());
 
-    expect_app_state_equal(final_state, expected_state);
+    expect_design_equal(final_state.design, expected_design);
```
dave-doty, line 1974: These tests aren't ideal; e.g. this line is testing the entire design when all that changed was to remove helices. Ideally you'd just test the set of helices to ensure it changed as you expect (without also testing strands, for example). But it's not worth nitpicking for now.

dave-doty (approved):

test/reducer_test.dart
UnHumbleBen: I'm kinda on the fence about this. On the one hand, I do see what you mean about not wanting to assert checks for the entire design because yes we want our unit tests to test for specific functionality and not worry about asserting checks on other parts of the design. On the other hand, there is some robustness in assuring that not only did a specified change happen, but also checking that it is the **only** change that happened, which is what doing a full comparison of the designs allows for.

dave-doty: > I'm kinda on the fence about this. On the one hand, I do see what you mean about not wanting to assert checks for the entire design because yes we want our unit tests to test for specific functionality and not worry about asserting checks on other parts of the design. On the other hand, there is some robustness in assuring that not only did a specified change happen, but also checking that it is the **only** change that happened, which is what doing a full comparison of the designs allows for.

I see your point. Mostly I'm thinking of times when I've had to fix a unit test after implementing some new feature, but the fix was something incidental, and wasn't actually representing that something broke. It was more like, the definition of correct behavior changed, and the more unit tests there are that reference that behavior, the more fixing has to be done.

One example that springs to mind is the fact that SVG coordinates are stored in helices (which is a design flaw that I'd like to get rid of eventually). Sometimes unit tests would break because the SVG coordinates would change, but it was only because the unit tests hard-coded the old way of computing the SVG coordinates.

UnHumbleBen: I see what you mean. I did think it was strange for me to hard code the SVG values when I was writing those tests 😅. For this PR, I rewrote a few tests to get rid of comparing the svg values between helices. Going forward it might make sense to add some helper function that checks for equality between helices that ignores the svg position if we find ourselves needing to write unit tests for comparing helices that have non default svg positions.

PR #556, opened 2020-12-25
bumped minor version (though no breaking changes to the web interface) to keep with scadnano Python package, which added some breaking changes to the Python API

dave-doty:


dave-doty:
Release notes:

No changes, but bumping version to keep with Python package, which had some breaking changes; see _TODO link to Python release notes_.

PR #578, opened 2021-03-08
Fixes #536; copy strand details to clipboard as human-readable text

### LINKED ISSUE #536: copy strand details to clipboard as human-readable text
Feature request: Give a way to copy all strand info to the system clipboard in human-readable text format. This form of copying would not be needed so often (e.g. when discussing a design with a colleague over email it saves me typing out the name, or when sanity checking sequence against some other seq on my computer it saves me manually typing out the sequence), so I would personally prefer it to appear only in the 'right click' and 'Edit' menus as a second form of copy, additional to the standard notion.

UnHumbleBen:


dave-doty (changes requested):
1. Remove the separate menu item under Edit, and make this be the response to copying strands when the user presses Ctrl+C (or selecting "Copy" under the Edit menu). In other words, it should do everything it does now (to allow copy/pasting of strands within scadnano), and also it should trigger the middleware to alter the system clipboard to contain the text information about the strands.

2. How hard would it be to copy information about any selected objects? For example, DNA ends. This seems straightforward by modifying the middleware to iterate over all selected objects. If so, then the name of the Action should change to `CopySelectedObjectsToClipboard`, and similarly with the middleware.

3. In the middleware, change from the (quadratic-time and -memory) repeated concatenation of strings, to putting the strings in a list, then call the method `join` on the list at the end to concatenate them into one string. (which uses linear time and memory instead)

4. Once the above is implemented, add some documentation to the README, near wherever the copy/paste functionality is explained.

5. **Update:** Also, if we're going to bother with exporting JSON, we may as well make it a valid JSON object. So if exporting multiple items, surround it with list brackets, and separate the list items with commas (and also newlines to make it readable, i.e., called `list.join(',\n')`), so that it can be viewed in a JSON viewer easily.

UnHumbleBen:
> 2. How hard would it be to copy information about any selected objects? For example, DNA ends. This seems straightforward by modifying the middleware to iterate over all selected objects. If so, then the name of the Action should change to `CopySelectedObjectsToClipboard`, and similarly with the middleware.

DNA ends, insertions, deletions, and modifications do not have JSON representation, so the current method of using `json_encode()` does not work. Until we have a specific idea of how we want to represent the clipboard information for these objects, I think it is best to stick with only strands for now.

dave-doty (approved):

Issue #614, opened 2021-07-09
assign domain complement from bound domains if domains are selected (instead of whole strand)


cehsour:
An addition to issue #592 specifically for computing the complement from bound domains. The same context menu item does this action, but if domains are selectable, then individual selected/clicked domains are assigned names, instead of assigning to all domains on a strand.

dave-doty:
Has this issue been handled by PR #616?

I see that "closes #614" is in the title of that PR, but not any commit message. Be sure to put "closes #614" in a commit message. The GitHub action that automatically generates release notes gathers together all the commit messages, and this helps to identify which issues were closed and might require more explanation/documentation in the release notes.

See some releases for examples of how the commit messages get collected: https://github.com/UC-Davis-molecular-computing/scadnano/releases

If it has been handled, then add the label "closed in dev". The issue will be closed automatically once the changes make their way to the main branch. (So keep the issue open in GitHub.)

Issue #592, opened 2021-04-17
add "assign domain complement from bound strands" option


damienwoods:
Sometimes we build a structure by copy-pasting strands, in which case we may often use 
"assign DNA complement from bound strands", as in #518 . I'm requesting the same feature for domain names to save lots of manual editing of domain names.

dave-doty:
This means if domain `a` is bound to `b`, and `a.name="xyz"`, and `b.name` is undefined, then after using this feature, we would have `b.name="xyz*"`?

damienwoods:
yup

PR #615, opened 2021-07-13
#592 add assign domain complement from bound strands 

cehsour:


dave-doty (changes requested):
Does one of these commits close #614? If so that should be in the commit message. Or are you doing that on a separate branch with a separate PR?

cehsour:
@dave-doty No, we are doing that on a separate branch with a separate PR.

dave-doty (approved):

PR #616, opened 2021-07-14
Closes #614 add assign domain name complement from bound domains

### LINKED ISSUE #614: assign domain complement from bound domains if domains are selected (instead of whole strand)

rishabhmudradi:


lib/src/view/design_main_strand.dart
```diff
@@ -324,6 +334,15 @@ assigned, assign the complementary domain names sequence to this strand.
 ''',
           on_click: assign_domain_name_complement_from_bound_strands,
         ),
+        if (app.state.ui_state.select_mode_state.domains_selectable())
```
dave-doty, line 337: This should not be a separate menu item. Just use the original menu item, but use the Selectable information to determine whether to assign to selected domains, or selected strands. If domains are not selectable, then exactly the behavior for `assign_domain_name_complement_from_bound_strands` should happen. If domains are selectable, then the new behavior should be implemented.

dave-doty (changes requested):

lib/src/view/design_main_strand.dart
cehsour: We added a commit to implement this just now.

lib/src/view/design_main_strand.dart
```diff
@@ -189,15 +189,24 @@ class DesignMainStrandComponent extends UiComponent2<DesignMainStrandProps>
...
+      List<Strand> strands_selected = app.state.ui_state.selectables_store.selected_strands.toList();
+      if (!strands_selected.contains(props.strand)) {
+        strands_selected.add(props.strand);
+      }
+      var action = actions.AssignDomainNameComplementFromBoundStrands(strands_selected);
+      app.dispatch(action);
+    }
   }
 
   assign_domain_name_complement_from_bound_domains(Domain domain) {
```
dave-doty, line 212: As far as I can tell, the function `assign_domain_name_complement_from_bound_domains` is not being called anywhere. If that's the case, it should be removed.

dave-doty (changes requested):

dave-doty (dismissed):

PR #617, opened 2021-07-14
Fixes #613 fix bug where right click on DNAEnd raises exception

### LINKED ISSUE #613: fix bug where right-click on DNAEnd raises exception
Right click on a DNAEnd. This error appears:

design_main_strand.dart:354 Uncaught TypeError: Cannot read property 'name' of null
    at design_main_strand.dart:354
    at design_main_strand._$DesignMainStrandComponent.new.context_menu_strand (design_main_strand.dart:407)
    at design_main_strand_dna_end._$DesignMainDNAEndComponent.new.on_context_menu (design_main_strand_dna_end.dart:155)

rishabhmudradi:


lib/src/view/design_main_strand.dart
```diff
@@ -352,7 +352,7 @@ assigned, assign the complementary domain names sequence to this strand.
           title: 'set domain name',
           on_click: () => set_domain_name(domain),
         ),
-        if (domain.name != null)
+        if (domain?.name != null)
```
dave-doty, line 355: I agree this fixes the crash. However, I'm a bit confused under what conditions `domain` will be null. I'm worried that an earlier bug is causing it to be null when it should be not null, and we should be fixing that instead of checking for null here.

Could it be when a 5' or 3' end is clicked? If so, then that seems like a mistake. The component `DesignMainDNAEnd` as a `Domain` as a prop, so I think that domain should be passed as an argument when `context_menu_strand` is called in `on_context_menu` in design_main_strand_dna_end.dart.

Or is there some situation in which `context_menu_strand` is called when there is no `Domain` available associated with the click that caused it to be called?

dave-doty (changes requested):

lib/src/view/design_main_strand.dart
dave-doty: I fixed this in view/design_main_strand_dna_end.dart

dave-doty:
Note that I changed the title of this PR to match the issue it is resolving. It was difficult for me to find it earlier since the titles didn't match, when the PR was instead named after the technique used to originally fix the bug (which turned out not to be how we fixed it in the end).

Issue #618, opened 2021-07-17
move context menu position when it doesn't fit into window


sarahayu:
Ideally will be implemented with #557 
By default, the context menu should appear to the bottom-right of the cursor, but if it appears out of bounds of window it should be moved to the left/top of cursor. Submenus should be dealt with similarly

dave-doty:
> if it appears out of bounds of window it should be moved to the left/top of cursor

Or a hybrid. Perhaps a more flexible approach is to calculate where the bottom of the context menu will appear and then translate the menu up by the quantity `min(0, y_coordinate_of_context_menu_bottom - y_coordinate_of_viewport_bottom)`.

dave-doty:
Closed in 62da977

PR #624, opened 2021-07-21
closes #557 #618 enable submenus in context menus and top level menu and move context menu position when it doesn't fit into window

### LINKED ISSUE #557: enable submenus in context menus and top-level menu
The right-click context menu for Strand is getting long:

image

Implement submenus to make it easier to fit on the screen. For example, we can have submenus for:

- the four ways to reflect 
- setting/removing strand/domain names 
- editing DNA

It would also be useful to have this functionality available for the top-level menu (under File, Edit, View, etc.)

sarahayu:


lib/src/view/design_context_menu.dart
```diff
@@ -20,29 +22,152 @@ mixin DesignContextMenuProps on UiProps {
   ContextMenu context_menu;
 }
 
-class DesignContextMenuComponent extends UiComponent2<DesignContextMenuProps> with PureComponent {
+mixin DesignContextMenuState on UiState {
```
dave-doty, line 25: I'm a bit confused by this solution. We use React state very sparingly in scadnano (see https://github.com/UC-Davis-molecular-computing/scadnano/blob/main/CONTRIBUTING.md#unidirectional-data-flow-architecture-of-scadnano).

I think essentially the only place it is used is in dialog forms to interact with the user, where as they type or select options, we store what they've entered in React state, and only once they press OK do we dispatch an Action to change the Redux state, since this is more efficient than dispatching actions to update the entire Redux store every time they press a key or click a checkbox.

But in this case, I don't see why we'd want to store the width and height are stored as React state. They should not be changing dynamically after the context menu is created. They should just be calculated once when the context menu is created. Once it is created, it will have the same width and height until it goes away (i.e., the user selects an option, or they click elsewhere).

I'm also not sure what `Ref<DivElement>` is. Perhaps we can meet on Zoom and someone can explain that to me, and also explain why you think we need React state. Maybe we do and I'm just not seeing why, but from my perspective, everything we need to know about rendering the context menu is known at the time it is created, so it doesn't make sense why we'd want to have a mutable state and allow the width or height to change dynamically.

dave-doty (changes requested):

lib/src/view/design_context_menu.dart
dave-doty: On an unrelated note, when I run this locally, I see this error in the console:

image

Double-check the list of items returned from the menu `render` method to ensure that they have unique string "key" props (this is a prop that you don't declare in the props class, but you can pass it in when creating a component; see examples in the code). This is necessary whenever a React component's `render` method returns a list of items. It might be related to one of the new submenu components that was added.

dave-doty (approved):

dave-doty (approved):

PR #631, opened 2021-07-31
467 domain labels move when switching orientation of strand

cehsour:


lib/src/reducers/strands_reducer.dart
```diff
@@ -211,19 +211,26 @@ Strand move_strand(
       int delta_offset,
       bool delta_forward}) {
   List<Substrand> substrands = strand.substrands.toList();
+  var list_of_domains = [...strand.domains];
```
dave-doty, line 214: Is this a way to make a copy of the list `strand.domains`? Somehow this would seem more "canonical" to me (or at least it's more to me clear what it's doing):

```dart
var list_of_domains = List<Domain>.from(strand.domains);
```

The `[...list]` notation is intended to make it easier to include lists inline in a long list of items, like when specifying React components, e.g., 

```dart
return [
  first_component(),
  ...list_of_components(),
  last_component(),
]
```

Also, I tend not to use the [Hungarian notation](https://en.wikipedia.org/wiki/Hungarian_notation) unless there are two different types of collections with the same object and I want to know which is which, i.e., if both `domains_set` and `domains_list` were in the same scope. But when it's just one collection, I usually just use the plural such as `domains`. In particular, List is so much more common than Set that `domains` almost always for me refers to a list of domains.

lib/src/reducers/strands_reducer.dart
```diff
@@ -211,19 +211,26 @@ Strand move_strand(
...
     Substrand new_substrand = substrand;
     if (substrand is Domain) {
+
       num original_view_order = original_helices_view_order_inverse[substrand.helix];
       num new_view_order = original_view_order + delta_view_order;
       int new_helix_idx = current_group.helices_view_order[new_view_order];
       assert(new_helix_idx != null);
       Domain domain_moved = substrand.rebuild(
             (b) => b
+          ..name = list_of_domains[list_of_domains.length - counter - 1].name
```
dave-doty, line 233: I have a preference here to create a variable to refer to `list_of_domains[list_of_domains.length - counter - 1]` with an appropriate name, and then here just access its `name` field. (But I'm not sure which name to pick.)

dave-doty (changes requested):

dave-doty (approved):

Issue #608, opened 2021-06-17
Fix crash bug on assigning DNA sequence


cehsour:
The design raises an exception when right-clicking on the scaffold and assigning DNA from bound substrands.

```json
{
  "version": "0.16.0",
  "grid": "square",
  "helices": [
    {"grid_position": [0, 0], "max_offset": 48},
    {"grid_position": [0, 1], "max_offset": 48}
  ],
  "modifications_in_design": {
    "/5Biosg/": {
      "display_text": "B",
      "idt_text": "/5Biosg/",
      "display_connector": false,
      "location": "5'"
    }
  },
  "strands": [
    {
      "color": "#0066cc",
      "sequence": "AACGTAACGTAACGTAACGTAACGTAACGTAACGTAACGTAACGTAACGTAACGTAACGTAACGTAACG",
      "is_scaffold": true,
      "domains": [
        {"helix": 1, "forward": false, "start": 8, "end": 24, "deletions": [20]},
        {"helix": 0, "forward": true, "start": 8, "end": 40, "insertions": [[14, 1], [26, 2]]},
        {"loopout": 3},
        {"helix": 1, "forward": false, "start": 24, "end": 40}
      ]
    },
    {
      "color": "#f74308",
      "sequence": "ACGTTACGTTACGTTTTACGTTACGTTACGTT",
      "domains": [
        {"name": "XYZ", "helix": 1, "forward": true, "start": 8, "end": 24, "deletions": [20]},
        {"name": "ABC", "helix": 0, "forward": false, "start": 8, "end": 24, "insertions": [[14, 1]]}
      ]
    },
    {
      "color": "#57bb00",
      "sequence": "ACGTTACGTTACGTTACGCGTTACGTTACGTTAC",
      "domains": [
        {"helix": 0, "forward": false, "start": 24, "end": 40, "insertions": [[26, 2]]},
        {"name": "ABC*", "helix": 1, "forward": true, "start": 24, "end": 40}
      ],
      "5prime_modification": "/5Biosg/"
    }
  ]
}
```

PR #635, opened 2021-08-06
#608 fix crash bug on assigning dna sequence

cehsour:


test/assign_dna_unit_test.dart
```diff
@@ -0,0 +1,1202 @@
...
+            assign_complements: true,
+            dna_sequence: dna_sequence2,
+            warn_on_change: true,
+          ));
+      expect(state.design.strands[0].dna_sequence, 'TTTGCCCT');
+    });
+
+    test('AssignDNA__adapter_assigned_from_scaffold_and_tiles', () {
+      /*
+          0    5    10
```
dave-doty, line 546: Some of this ASCII art is a little tough to read. For instance, this one looks like the crossover from the top-most strand might be going down to the bottom-most strand, through the strand in the middle, but I'm not sure.

I know I said the convention doesn't matter as long as it's readable, but this isn't very readable to me. Maybe it would be better to use the convention from the Python comments in this case, where bases ACGT represent where DNA is, and dashes represent empty space, so that the crossover from the topmost strand to the bottom most strand is "visible" through the ASCII art. and doesn't "disappear behind" anything. (Or another convention that doesn't hide any crossovers.)

I didn't comment explicitly on any other ASCII art, but same applies to all of them.

dave-doty (changes requested):

dave-doty (approved):

Issue #628, opened 2021-07-26
Deleting Helices when using multiple groups of helixes causes the saved .sc file to be an "Illegal Design" when opened


caiwood:
When working with multiple groups of helixes, if one deletes helixes from a group (e.g helixes 80 and 81 in the below error message) and saved as a .sc file, then upon opening the file in scadnano, the following error is returned:

Error Message:
******************
* illegal design *
******************

The design has the following problem:

number of helices (12) does not match length of helices_view_order (14)
helix idxs:         68, 69, 70, 71, 72, 73, 74, 75, 76, 77, 78, 79
helices_view_order: 68, 69, 70, 71, 72, 73, 74, 75, 76, 77, 78, 79, 80, 81

**********************************************************************************
* If you believe this is due to a bug in scadnano, please file a bug report at   *
*   https://github.com/UC-Davis-molecular-computing/scadnano/issues              *
* Include this entire message in the email.                                      *
**********************************************************************************

stack trace:
Instance of 'IllegalDesignError'
    at Object.wrapException (https://scadnano.org/main.dart.js:1181:17)
    at Object.HelixGroup_from_json (https://scadnano.org/main.dart.js:32127:19)
    at Object.Design__groups_from_json (https://scadnano.org/main.dart.js:37534:19)
    at Object.Design__helices_and_groups_from_json (https://scadnano.org/main.dart.js:37671:32)
    at Object.Design_from_json (https://scadnano.org/main.dart.js:37781:13)
    at Object.load_dna_file_reducer (https://scadnano.org/main.dart.js:23252:40)
    at StaticClosure.app_state_reducer (https://scadnano.org/main.dart.js:30360:18)
    at Store__createReduceAndNotify_closure.call$1 (https://scadnano.org/main.dart.js:66087:28)
    at StaticClosure.oxdna_export_middleware (https://scadnano.org/main.dart.js:36842:12)
    at Store__createDispatchers_closure.call$1 (https://scadnano.org/main.dart.js:66097:34)
-----------------------------------------------------
End of Error Message:

The file can be manually fixed by removing the deleted helixes from the "Helices_view_order" for the relevant group at the top of the .sc file. 

The Zip file contains both the broken .sc file as well as the manually fixed file.
[ScadnanoBugFilesHelixDeletion.zip](https://github.com/UC-Davis-molecular-computing/scadnano/files/6933528/ScadnanoBugFilesHelixDeletion.zip)

[...]

PR #636, opened 2021-08-11
#628 deleting helices when using multiple groups of helixes causes the saved .sc file to be an illegal design when opened

cehsour:


dave-doty (approved):
I have to be honest, I'm not sure what I was thinking with the old code or how I expected that to work. This looks much better.

dave-doty:
Is there a commit that will automatically close the issue when this is pushed to main? If not, let's create one. Also remember to set closed_in_dev on the relevant issue.

cehsour:
No sorry, I will revert the pull request and add the closes commit.

Issue #561, opened 2020-12-28
add ability to edit IDT fields of a strand


dave-doty:
Currently, IDT fields can only be edited in the [Python scripting library](https://scadnano-python-package.readthedocs.io/en/latest/#scadnano.Strand.idt) manually in the `.sc` file. Give an interface to edit them in the web interface.

At the same time, implement issue #560, so that the IDT fields will no longer include a name (instead, the exported name is `Strand.name`).

dave-doty:
I'd recommend figuring out issue #557 first, since this will add even more options to the right-click context menu.

dave-doty:
Go here for scale and purification options: https://www.idtdna.com/site/order/oligoentry

Have a dropdown list with all of these, and an "Other..." option to type a custom value.

PR #643, opened 2021-08-24
#561 add ability to edit idt fields of a strand 

rishabhmudradi:


lib/src/state/dialog.dart
```diff
@@ -25,6 +25,7 @@ abstract class Dialog with BuiltJsonSerializable implements Built<Dialog, Dialog
       {String title,
       Iterable<DialogItem> items,
       Iterable<Iterable<int>> mutually_exclusive_checkbox_groups = const [],
+      Iterable<int> force_disable = const {},
```
dave-doty, line 28: I don't quite understand this `force_disable` option. Maybe it's needed, but if we can use the existing framework to enable/disable certain input fields, that's preferable. Maybe we can talk about on Wednesday why this is needed. (Or just explain in a reply here.)

dave-doty (changes requested):
This is probably okay, but I'm feeling lazy and it would be easier if someone just walks me through the new functionality on Wednesday rather than me testing it myself right now. :)

dave-doty (approved):
If this is all the changes we discussed at the last meeting, then it's good, but I haven't really checked carefully.

PR #644, opened 2021-08-24
fix shift-clicking not working at times

sarahayu:


lib/src/view/design.dart
```diff
@@ -364,6 +364,42 @@ class DesignViewComponent {
...
+            app.dispatch(action_adjust);
+          }
+
+          var action_remove = actions.SelectionRopeRemove();
+          app.dispatch(action_remove);
+        }
+      }
+    }
+
+    window.onBlur.listen((_) => end_select_mode());
```
dave-doty, line 401: I just want to double-check: other than moving some code into the new function `end_select_mode`, this line with `window.onBlur.listen` is the only real change to the logic, right?

dave-doty (approved):

lib/src/view/design.dart
sarahayu: Yes, that's correct

Issue #403, opened 2020-07-21
add axis arrows to main view and side view


dave-doty:
The conventions for interpreting geometry in the side view and main view are a bit convoluted, and buried somewhere in the README text. It is particularly tricky if "invert y axis" is selected, since the z axis must also be inverted, or the perspective must be switched, so that now we imagine the side view is looking from right to left in the main view (instead of from left to right as it normally is).

It would be easier to follow what's going on if these ideas were explicitly displayed in the side view and main view, using cartesian axis arrows, and possibly some representation somewhere of what the side view represents relative to the main view (i.e., an eye looking in one direction left or right at a depiction of a design).

rishabhmudradi:
Hi Professor, just wanted to confirm that you wanted us to remove the "eye" portion of this feature.

dave-doty:
> Hi Professor, just wanted to confirm that you wanted us to remove the "eye" portion of this feature.

Yes, it's not particularly clear what it means, and I can't think of a way to make it clearer.

PR #646, opened 2021-09-07
#403 add axis arrows to main view and side view

cehsour:


dave-doty (changes requested):
I noticed a bug (not sure if it's related to this issue, but I think it's only on this branch). The side view helices overlap the bottom of the view a bit too much:

image

dave-doty (approved):

PR #649, opened 2021-09-19
fixes #645: crash on edit modification

### LINKED ISSUE #645: crash on edit modification
Create a 5' modification and right-click to edit it:

image

rishabhmudradi:


lib/src/view/design_main_strand.dart
```diff
@@ -730,8 +728,16 @@ PAGEHPLC : Dual PAGE & HPLC
 
   Future<void> ask_for_add_modification(
       [Domain domain = null, Address address = null, bool is_5p = null]) async {
-    assert((is_5p == null && domain != null && address != null) ||
-        (is_5p != null && domain == null && address == null));
+    /*
```
dave-doty, line 731: There should be more documentation here. Specifically, due to how `context_menu_strand` is called, there's some convention for which of these can be null and which aren't. Document it.

For example, I suspect that if one right-clicks on a domain (not a 5' or 3' end), then is_5p is null.

Also, please change `is_5p` from type bool to an enumerated type (not just here, but back where this data gets initialized, when `context_menu_strand` is called) called `ModificationType` with three possible values: `five_prime`, `three_prime`, and `internal`:

```dart
enum ModificationType { five_prime, three_prime, internal }
```

Then pass one of those values as input here (`internal` if the right-click was on a domain, for instance). Put that enumerated type in state/modification.dart.

dave-doty (changes requested):

dave-doty (commented):
Can we walk through these changes at the next meeting? It's been a while since I've thought about this issue.

dave-doty (approved):

dave-doty (approved):

Issue #224, opened 2020-04-20
allow edit mode/selection mode to be hidable, and replace words with small icons


dave-doty:
Allow the menu of edit modes/selection modes on the right to be hidable, and change them as in cadnano to be represented by a small icon, with the text only appearing on a hover.

But, keep the shortcut key visible without a hover. (e.g., 'S' for "Select").

PR #651, opened 2021-09-25
#224 allow edit mode/selection mode to be hidable and replace words with small icons

rishabhmudradi:


dave-doty (changes requested):
For insertion, deletion, modification, scaffold, and staple, let's just make it one domain instead of two. So for each, get rid of the crossover and the bottom domain. This will save some height and I think communicate just as well what it means.

image

For domain, show both domains, but show only the top one selected.

image

For crossover, make the selected crossover slightly darker (maybe not as dark as a domain but more than currently), it's a bit hard to see

image

dave-doty (changes requested):
Can we put the SVG files in two subdirectories of images/, named images/edit_mode_icons and images/selectable_icons? This will help keep them organized and separate from the other images.

Issue #671, opened 2021-11-08
fix crash when loading for first time


dave-doty:
Clear the browser's local storage and start up scadnano. The following exception occurs:

```
(index):388 Uncaught (in promise) TypeError: Cannot read properties of undefined (reading 'setZoomScaleSensitivity')
    at set_zoom_speed ((index):388)
    at app.App.new.start (app.dart:98)
    at start.next (<anonymous>)
    at runBody (dart_sdk.js:35775)
    at Object._async [as async] (dart_sdk.js:35806)
    at app.App.new.start (app.dart:80)
    at main$ (main.dart:5)
    at <anonymous>:1:8
    at Object.runMain (client.js:7952)
    at client.js:23849
    at _wrapJsFunctionForAsync_closure.$protected (client.js:3003)
    at _wrapJsFunctionForAsync_closure.call$2 (client.js:10800)
    at Object._asyncStartSync (client.js:2968)
    at main__closure1.$call$body$main__closure (client.js:23861)
    at main__closure1.call$1 (client.js:23792)
    at StaticClosure._rootRunUnary [as call$2$5] (client.js:3303)
    at _CustomZone.runUnary$2$2 (client.js:12130)
    at _CustomZone.runUnaryGuarded$1$2 (client.js:12058)
    at _ControllerSubscription._sendData$1 (client.js:11601)
    at _DelayedData.perform$1 (client.js:11754)
    at _PendingEvents_schedule_closure.call$0 (client.js:11807)
    at Object._microtaskLoop (client.js:3168)
    at StaticClosure._startMicrotaskLoop (client.js:3174)
    at _AsyncRun__initializeScheduleImmediate_internalCallback.call$1 (client.js:10675)
    at invokeClosure (client.js:1188)
    at MutationObserver.<anonymous> (client.js:1207)
```

Issue #670, opened 2021-11-08
fix bug where edit mode menu hidden on zoom in


dave-doty:
Use the browser's Zoom (Ctrl++) to Zoom in above 100% on the whole page, for example, in Chrome:

![image](https://user-images.githubusercontent.com/19274365/140772142-f4bb8d41-735d-4c0d-bea2-848fd37fb86a.png)

Once Zoomed in far enough, the Edit mode is hidden and only Selectables display. This is the far right side of the screen:

![image](https://user-images.githubusercontent.com/19274365/140772289-232d06ba-1f99-42eb-b547-ca03209294df.png)

PR #672, opened 2021-11-11
#671 fix crash when loading for first time

cehsour:


cehsour:
Fixed both 671 and 670 issues

dave-doty (changes requested):
Can you add a commit(s) that includes the title of the issue being fixed, for example

fixes #670: edit mode menu hidden on zoom in
fixes #671: fix crash when loading for first time

That way it automatically makes its way into the release notes under the "Commits" menu: https://github.com/UC-Davis-molecular-computing/scadnano/releases

dave-doty (approved):

Issue #674, opened 2021-11-11
remove warning that appears in console on local host when adding modification


cehsour:
<img width="1680" alt="Screen Shot 2021-11-11 at 11 09 11 AM" src="https://user-images.githubusercontent.com/84648681/141354969-e3a7d1a9-613f-4711-9784-7e62255b16a4.png">

PR #675, opened 2021-11-11
#674 remove warning that appears in console on local host when adding modification

sarahayu:


lib/src/serializers.dart
```diff
@@ -324,7 +324,16 @@ Serializers serializers = _$serializers;
 Serializers standard_serializers = (serializers.toBuilder()
       ..add(PointSerializer<num>())
       ..add(ColorSerializer())
-      ..addPlugin(new StandardJsonPlugin()))
+      ..addPlugin(new StandardJsonPlugin())
```
dave-doty, line 327: Can you add a link to a webpage explaining this? (assuming there is one that suggested this fix)

dave-doty (changes requested):

dave-doty (approved):

Issue #679, opened 2021-11-14
move selected helices to group


dave-doty:
If a user selects some helices in the side view:

![image](https://user-images.githubusercontent.com/19274365/141693135-8a97993d-6f28-41c5-ad76-29eb42552168.png)

They should be able to right-click and see an option "Move to another helix group" that moves them all to a different helix group.

dave-doty:
This has been implemented in 3b35ac8 but needs unit tests.

dave-doty:
TODOs:

- add unit tests
  - moving from multiple helix groups
  - overlapping helices should not be allowed to move
  - verify helices_view_order is correct in from and to groups
- The second above needs to be fixed with a middleware check that pops up `window.alert` to tell the user to move helices out of the way in the "to" group.

dave-doty:
Bug

Here's a bug. Load this design

[tbn-detector-3x3-pre.zip](https://github.com/UC-Davis-molecular-computing/scadnano/files/7565179/tbn-detector-3x3-pre.zip)

Select all helices in group "depth 2 untriggered". Move them to group "depth 1 untriggered".

Bug #1: They cover helices 0,1,2, but should be displayed below them. (May be an instance of #677).

Bug #2: Reload the page. You get this error (showing helices_view_order was set improperly, a bug probably hidden by the first bug): Actually I couldn't reproduce the following error, but it happens when loading this design, which you get by saving the design after doing the move group above:  [tbn-detector-3x3.zip](https://github.com/UC-Davis-molecular-computing/scadnano/files/7565206/tbn-detector-3x3.zip)

```
******************
* illegal design *
******************

The design has the following problem:

number of helices (7) does not match length of helices_view_order (10)
helix idxs:         0, 1, 2, 4, 5, 6, 7
helices_view_order: 0, 1, 2, 4, 5, 6, 7, 0, 1, 2

**********************************************************************************
* If you believe this is due to a bug in scadnano, please file a bug report at   *
*   https://github.com/UC-Davis-molecular-computing/scadnano/issues              *
* Include this entire message in the email.                                      *
**********************************************************************************

stack trace:
dart:sdk_internal 4850:11                                                                         throw_
package:scadnano/src/state/group.dart 101:9                                                       from_json
package:scadnano/src/state/design.dart 1020:79                                                    _groups_from_json
package:scadnano/src/state/design.dart 1099:9                                                     _helices_and_groups_from_json
[...]

Issue #667, opened 2021-11-07
make scadnano pitch angle agree with oxDNA


dave-doty:
**Note:** This is a breaking change since it will change how oxDNA output works.

Currently, scadnano interprets the pitch angle as a clockwise rotation in the Y-Z plane, following SVG convention. The following design has a helix group (containing helix 1) with pitch=45 (clockwise, away from the single strand on helix 0):

![image](https://user-images.githubusercontent.com/19274365/140661971-de0ef47d-e112-464c-b49c-34003d7b43c9.png)

However, exporting to oxDNA rotates the helix in the opposite direction (counter-clockwise, towards the single strand on helix 0):

![image](https://user-images.githubusercontent.com/19274365/140661970-f1d2367a-7adb-4bd1-aa71-9cc290226dac.png)

The two conventions should match, either by rotating counter-clockwise in the scadnano main view, or by changing the oxDNA export code to rotate clockwise in the Y-Z plane. **UPDATE:** We changed the oxDNA export to rotate clockwise in the Y-Z plane.

UnHumbleBen:
@dave-doty I am working on the release note and have some questions:

Firstly, which option was used to resolve this issue? You discussed two options here but which one was chosen?
> The two conventions should match, either by rotating counter-clockwise in the scadnano main view, or by changing the oxDNA export code to rotate clockwise in the Y-Z plane.

Secondly, I would like to take a screenshot of either scadnano or oxDNA after the fix. Can I have the scadnano or oxDNA file generated after the fix? Or at the very least, the scadnano file so I can generate the oxDNA file myself?

dave-doty:
> The two conventions should match, either by rotating counter-clockwise in the scadnano main view, or by changing the oxDNA export code to rotate clockwise in the Y-Z plane.

We chose the second, altering the oxDNA export to rotate clockwise in the Y-Z plane.

> Secondly, I would like to take a screenshot of either scadnano or oxDNA after the fix. Can I have the scadnano or oxDNA file generated after the fix? Or at the very least, the scadnano file so I can generate the oxDNA file myself?

I don't think I saved it. :) But it was fairly straightforward to make.

UnHumbleBen:
> I don't think I saved it. :) But it was fairly straightforward to make.

[...]

Issue #665, opened 2021-11-07
export oxDNA for only selected strands


dave-doty:
Similar to #537.

PR #682, opened 2021-11-28
0.17.0

UnHumbleBen:


dave-doty:
Document these in release notes:

#667
#679
#561
#403
#224
#665

lib/src/reducers/helices_reducer.dart
```diff
@@ -548,7 +549,7 @@ BuiltMap<int, Helix> helix_position_set_reducer(
 /// are changing in the action, then state will not yet have been updated to the new value; in that case
 /// assign them explicitly in the parameter list; e.g., see [invert_xy_set_helices_reducer]
```
dave-doty, line 550: Can you also adjust the comment above this function to mention `groups` as well?

dave-doty: More generally, can we document what seems to be true, that this function is a wrapper for `util.helices_assign_svg`, which allows one to avoid listing the various parameters that function needs, if they can simply be found in a single instance of `AppState`, but that these optional parameters are necessary when some parts of the `AppState` instance aren't equal to what is needed in `util.helices_assign_svg`? Can you also add a comment above `util.helices_assign_svg` indicating the same (it has a long parameter list; if those parameters can be found in an instance of `AppState` then call `reassign_svg_positions` instead).

dave-doty (changes requested):

dave-doty (approved):

dave-doty (approved):

UnHumbleBen:
@dave-doty I finished drafting up the release note, let me know if there's anything you would like to change (or you can edit it yourself).

Additionally, I found a bug that should be addressed before merging to main as this bug only appears on dev and is breaking "Move Group" which is something that is already working in main: #696

dave-doty (approved):

dave-doty:
What happened to the automated release notes generator? It should have put something here:

https://github.com/UC-Davis-molecular-computing/scadnano/releases

PR #685, opened 2021-11-30
Work In Progress: Fixes #677; adjusting helices view order should update view

### LINKED ISSUE #677: adjusting helices view order should update view
Click on Group&rarr;adjust current group, and change the helices view order. It does not update the view. It does save the change in the design so that if you refresh the page, the new view order is displayed.

UnHumbleBen:
Added a unit test to test for this scenario

### DIFF
--- a/test/reducer_test.dart
+++ b/test/reducer_test.dart
@@ -7534,16 +7534,19 @@ main() {
     Helix helix1 = Helix(idx: 1, grid_position: GridPosition(0, 1), group: "foo");
     HelixGroup group = HelixGroup(helices_view_order: [0,1], grid: Grid.square);
     Design design = Design(helices: [helix0, helix1], groups: {"foo": group});
+    helix0 = design.helices[0];
+    helix1 = design.helices[1];
     AppState state = app_state_from_design(design);
-    Point<num> original_helix0_svg_position = state.design.helices[0].svg_position;
+    // Point<num> original_helix0_svg_position = state.design.helices[0].svg_position;
     Point<num> original_helix1_svg_position = state.design.helices[1].svg_position;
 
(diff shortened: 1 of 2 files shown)


UnHumbleBen:
@dave-doty Just want to share the test case I have written and get your opinion on it.

test/reducer_test.dart
```diff
@@ -7527,6 +7527,22 @@ main() {
...
+    HelixGroup group = HelixGroup(helices_view_order: [0,1], grid: Grid.square);
+    Design design = Design(helices: [helix0, helix1], groups: {"foo": group});
+    AppState state = app_state_from_design(design);
+    Point<num> original_svg_position = state.design.helices[0].svg_position;
+
+    HelixGroup new_group = HelixGroup(helices_view_order: [1,0], grid: Grid.square);
+    AppState new_state = app_state_reducer(state, GroupChange(old_name: "foo", new_name: "foo", new_group: new_group));
+
+    // New svg position should have changed
+    expect(new_state.design.helices[0].svg_position == original_svg_position, false);
```
dave-doty, line 7544: I'm a bit confused. Shouldn't we be testing that `new_state.design.helices[1].svg_position` is equal to `original_svg_position`, since helices 0 and 1 should swap positions in this change?

Maybe we can rename the variable `original_svg_position` to `original_helix0_svg_position` to emphasize which helix it's referring to.

dave-doty (changes requested):

test/reducer_test.dart
UnHumbleBen: Good point. I wasn't sure if the new svg positions would necessarily be just the old ones but swapped but considering how svg_positions are by default calculated from the view order and the view order has been swapped, then of course we should expect the svg position to swap. I have made the change, will work on implementing the fix now.

dave-doty (approved):

PR #703, opened 2021-12-22
Work in Progress: Fixes #650; move svg position of helices out of design and into app state

### LINKED ISSUE #650: move svg position of helices out of design and into app state
That was, for instance, dispatching an action such as InvertXYSet, which does not change anything abou the design that would get stored in the `.sc` file, will not change the Dart `Design` object and cause a warning about changing the `Design` with a non-`UndoableAction` action.

UnHumbleBen:
Drafting a PR so that progress can be reviewed if needed

Work in Progress: Fixes #650; move svg position of helices out of design and into app state

### DIFF
--- a/lib/src/actions/actions.dart
+++ b/lib/src/actions/actions.dart
@@ -705,17 +705,17 @@ abstract class SetOnlyDisplaySelectedHelices
 /////////////////////////////////////////////////////////////////////////////////////////////////////////////
 // invert y-axis
 
-abstract class InvertXYSet
+abstract class InvertYSet
     with BuiltJsonSerializable
-    implements Action, SvgPngCacheInvalidatingAction, Built<InvertXYSet, InvertXYSetBuilder> {
-  bool get invert_xy;
+    implements Action, SvgPngCacheInvalidatingAction, Built<InvertYSet, InvertYSetBuilder> {
+  bool get invert_y;
 
   /************************ begin BuiltValue boilerplate ************************/
-  factory InvertXYSet({bool invert_xy}) = _$InvertXYSet._;
+  factory InvertYSet({bool invert_y}) = _$InvertYSet._;
 
-  InvertXYSet._();
+  InvertYSet._();
 
-  static Serializer<InvertXYSet> get serializer => _$invertXYSetSerializer;
+  static Serializer<InvertYSet> get serializer => _$invertYSetSerializer;
 }
 
 /////////////////////////////////////////////////////////////////////////////////////////////////////////////
--- a/lib/src/middleware/autostaple_and_autobreak.dart
+++ b/lib/src/middleware/autostaple_and_autobreak.dart
@@ -53,7 +53,7 @@ _autobreak(Store<AppState> store, actions.Autobreak action) async {
 void _handle_response(Store<AppState> store, http.Response response) {
   if (response.statusCode == 200) {
     var json_model_text = response.body;
-    var design_new = Design.from_json_str(json_model_text, store.state.ui_state.invert_xy);
+    var design_new = Design.from_json_str(json_model_text, store.state.ui_state.invert_y);
     store.dispatch(actions.NewDesignSet(design: design_new));
   } else {
     Map response_body_json = jsonDecode(response.body);
--- a/lib/src/reducers/app_ui_state_reducer.dart
+++ b/lib/src/reducers/app_ui_state_reducer.dart
@@ -160,7 +160,7 @@ bool show_mismatches_reducer(bool _, actions.ShowMismatchesSet action) => action
 bool show_domain_name_mismatches_reducer(bool _, actions.ShowDomainNameMismatchesSet action) =>
     action.show_domain_name_mismatches;
 
-bool invert_xy_reducer(bool _, actions.InvertXYSet action) => action.invert_xy;
+bool invert_y_reducer(bool _, actions.InvertYSet action) => action.invert_y;
 
 bool warn_on_exit_if_unsaved_reducer(bool _, actions.WarnOnExitIfUnsavedSet action) => action.warn;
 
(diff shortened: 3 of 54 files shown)


UnHumbleBen:
@dave-doty I have completed the refactoring. I tried loading very_large_origami design and performing these actions:
* moving strands around
* changing view order
* inverting xy

No performance drop detected for these actions. Are there any other actions/situations you want me to test out that could potentially have performance concerns?

dave-doty:
I'm seeing OverReact JSON warnings (`Converting object to an encodable object failed: Instance of '_$36LoadDNAFile'`), but I assume this is fixed in the dev branch. Just double-checking to make sure it's not a new error.

dave-doty:
> No performance drop detected for these actions. Are there any other actions/situations you want me to test out that could potentially have performance concerns?

I agree, it seems to be fairly fast. But if you put a print statement in `helices_assign_svg`, you can see that it is being re-calculated on every single dispatch of an Action to the main store. (The optimized stores, such as those for drawing the selection rectangle or dragging DNA ends, do not result in re-calculating it.)

I suppose it's fine as long as it's fast, but I still wish we could figure out a way to set it up so it's automatically re-calculated, but only when really needed. Right now it gets re-calculated on every single update to AppState (which is essentially every action).

At the very least, we should optimize the implementation of `helices_assign_svg` so that it's as fast as possible since it is now getting called so often.

dave-doty:
> At the very least, we should optimize the implementation of `helices_assign_svg` so that it's as fast as possible since it is now getting called so often.

I saw one optimization: Now that there's only one call to `helices_assign_svg` (outside of unit tests), there's no point in having it accept a `Map` for `helices` since what we have is a `BuiltMap`.

I also added a print statement in the memoized field `AppState.helix_idx_to_svg_position_map` to show how many microseconds it takes. I created a design with 100 helices and test it by moving a strand around.

Currently it takes between 500 and 800 microseconds per call for a design with 100 helices:

image

That's not bad, but it might be a good idea to see if there's anything else that can be optimized.

dave-doty (changes requested):
See comments above.

dave-doty:
It also seems that while we're bothering to refactor in this way, perhaps `Helix.invert_xy` should be removed also. (It also seems like an outdated name since it refers to the UI state `invert_yz` property.)

I thought about also saying we should remove `Helix.geometry`, but I don't think we should, since `Helix.geometry` is used in many more places and is used for things other than calculating the SVG position, for instance in translating grid position to 3D position. So it's nice to have it available in each Helix.

dave-doty:
There's a bug. Load [this design](https://github.com/UC-Davis-molecular-computing/scadnano/files/7799369/4-helices.zip)
and select helices 0 and 2:

image

Then select View --> Display only selected helices. The expected behavior is for helices 0 and 2 to be as close in the main view as 0 and 1 are above, but instead they are spaced:

image

On the dev branch the behavior is as expected:

image

The problem is that `helices_assign_svg` is not given access to the field `AppUIState.only_display_selected_helices`. That should be passed into `helices_assign_svg`, or possibly used prior to calling `helices_assign_svg` in the memoized getter `helix_idx_to_svg_position_map` in order to determine how to set `selected_helix_idxs`.

dave-doty:
**TODO:** Sorry I was lazy and didn't do this myself. Can we add a unit test to reproduce the above bug related to displaying only selected helices?

UnHumbleBen:
**Update**: Removed `Helix.invert_y`

Now will add display selected helices unit test as discussed here https://github.com/UC-Davis-molecular-computing/scadnano/pull/703#issuecomment-1003740603

UnHumbleBen:
@dave-doty 

Ideally, svg positions will only be recalculated on updates to the parameters to `util.helices_assign_svg`. I was wondering if it was possible for `@memoized` to be enhanced so that it would only recompute fields if the specific necessary fields for the memoization have been changed, but based on https://github.com/google/built_value.dart/issues/774, it seems like there are some issues with this feature. It's been a while since I read that issue so I will need to reread it.

I will also think about optimizations we can make for the `util.helices_assign_svg` function itself, but I feel like the main issue is that we are calling this function too much rather than the runtime of the function itself.

dave-doty:
> I was wondering if it was possible for `@memoized` to be enhanced so that it would only recompute fields if the specific necessary fields for the memoization have been changed, but based on [google/built_value.dart#774](https://github.com/google/built_value.dart/issues/774), it seems like there are some issues with this feature.

That issue was suggesting a much less sophisticated approach that would not help here. In that issue, I was suggesting that the memoized field be copied if *no* fields in the same class changed. 

So even if that feature were implemented, it wouldn't help us here, because each time AppState is updating, some field in it is actually changed, just usually not one that affects `AppState.helix_idx_to_svg_position_map`.

It would be much more difficult (undecidable, in fact) to automatically detect which specific fields the memoized field depends on.

Of course, one could imagine an approach where the user annotates `@memoized` with a list of fields it depends on, but this would have the same potential for bugs: if one changes the code and now the memoized field has a new dependency, but you forget to add it to the list, then the field will not be properly updated.

The only other (ugly) idea I had was to re-structure the AppState tree like this:

AppState
- AppUIState
- DesignAndOthers
  - Design
  - invert_y
  - selected_helix_idxs
  - helix_idx_to_svg_position_map

In other words, pull the specific parts of the AppUIState that affect `helix_idx_to_svg_position_map` into a subtree DesignAndOthers that's separate from the rest of the AppUIState. Then `helix_idx_to_svg_position_map` is only updated when some part of DesignAndOthers changes. 

This is very ugly and counterintuitive, so I don't want to do this since the performance doesn't seem to be bad right now.

UnHumbleBen:
I see, I didn't realize that that it would be undecidable to detect which fields are necessary. Kinda curious of a proof of this. Is it similar to the proof for halting problem undecidability?

The idea of annotating the @memoized field with the specific field dependency, but I agree, this would also present another source for bugs.

Can't think of any other ideas for how to optimize the function being called, so I'll focus on optimizing the function

UnHumbleBen:
**Update**: I was able to achieve a 2x speedup for a 100 helices design by removing unnecessary data structures and sorting. I'm not sure how to optimize the function further. Do you think this is good enough to merge to dev?

dave-doty:
> I see, I didn't realize that that it would be undecidable to detect which fields are necessary. Kinda curious of a proof of this. Is it similar to the proof for halting problem undecidability?

Suppose I could decide whether memoized field `b` depends on field `a`, i.e., whether changing `a` can change the value of `b`, using a routine `a_depends_on_b(C)`, where `C` is the class with fields `a` and `b`. Then we can decide the halting problem (does `M` halt on the empty string?) as follows:

```python
def halts(M):
    @dataclass
    def P:
        a: int
        
        @getter
        def b(self):
            M("")
            return a

    return a_depends_on_b(C)
```

if `M("")` halts, then `C.b` is always equal to `C.a` (thus depends on `C.a`), otherwise `C.b` is always undefined (thus does not depend on `C.a`).

**Challenge:** adjust this proof so that `C.b` always has a defined value, even if `M("")` does not halt.

dave-doty:
> **Update**: I was able to achieve a ~2x speedup for a 100 helices design by removing unnecessary data structures and sorting. I'm not sure how to optimize the function further. Do you think this is good enough to merge to dev?

Great! Sure, we can merge now.

UnHumbleBen:
@dave-doty Seems like a suggested change was requested that has not been resolved so requested another review just in case there's something I missed.

dave-doty (approved):

PR #713, opened 2022-01-04
closes #711: remember previously chosen M13 rotation in Assign DNA sequence dialog

### LINKED ISSUE #711: remember previously chosen M13 rotation in Assign DNA sequence dialog
After opening the 'assign DNA sequence' dialog, for the default for 'rotation of predefined DNA sequence:' I see 5587, even though the design I'm working on has another rotation already set (5514). I think its best to have the values from the currently displayed design as defaults, otherwise the user could unwittingly make a major design change by leaving the defaults as they are and clicking 'OK'.

dave-doty:


dave-doty:
Ben, I changed stuff around a bit for how we store DNA Assignment options in the state and action that is dispatched when assigning DNA. I tested it and the unit tests passed, but can you sanity check it?

UnHumbleBen:
Sure, I'll get to it tonight or tomorrow morning

lib/src/view/design_main_strand.dart
```diff
@@ -971,16 +962,18 @@ Future<void> ask_for_assign_dna_sequence(
   List<DialogItem> items = [null, null, null, null, null, null, null];
 
   items[idx_sequence] =
-      DialogTextArea(label: 'sequence', value: strand.dna_sequence ?? '', rows: 10, cols: 80);
-  items[idx_use_predefined_dna_sequence] = DialogCheckbox(label: 'use predefined DNA sequence');
+      DialogTextArea(label: 'sequence', value: strand.dna_sequence ?? '', rows: 4, cols: 80);
```
UnHumbleBen, line 965: Question: what's the intention of changing the number of rows from 10 to 4?

lib/src/view/design_main_strand.dart
```diff
@@ -997,18 +990,20 @@ Future<void> ask_for_assign_dna_sequence(
 
   String dna_sequence;
 
+  int m13_rotation = options.m13_rotation;
```
UnHumbleBen, line 993: nit: `m13_rotation` is only used inside of the body of the if statement, so I would declare it inside of the if statement.

UnHumbleBen (approved):
Looks good to me, just a minor nitpick on a variable declaration placement and a question

lib/src/view/design_main_strand.dart
dave-doty: > Question: what's the intention of changing the number of rows from 10 to 4?

On my laptop screen, 10 rows pushes the dialog off the bottom of the page.

lib/src/view/design_main_strand.dart
dave-doty: It's used again near the end of the function.

PR #723, opened 2022-01-14
closes #719: add "reset local settings" button

### LINKED ISSUE #719: add "reset local settings" button
Sometimes it is possible to fix a problem with scadnano by deleting the brower's local storage.

Add a button that lets the user do this without going through developer tools.

**UPDATE:** Actually let's do two buttons, under File: 

1. Delete stored design.
2. Reset local settings.

Also, if an error message is being shown (the one that says "You have discovered a bug in scadnano"), that message should always include a suggestion to press File --> Delete the stored design.

rishabhmudradi:


lib/src/view/menu.dart
```diff
@@ -222,6 +223,16 @@ really want to exit without saving.'''
         ..display = 'Import cadnano v2'
         ..key = 'import-cadnano')(),
       DropdownDivider({'key': 'divider-import-cadnano'}),
+      (MenuDropdownItem()
```
dave-doty, line 226: I'd prefer if we did this the "canonical" React/Redux way of using middleware. Writing to local storage is a side effect (i.e., it modifies something outside of the Redux State), and the traditional way to do side effects is by dispatching an Action and then intercepting the Action in middleware.

So I think we should make a new Action called `ResetLocalStorage` and dispatch it from this menu click. Then write a new middleware (in a new file in the middleware directory) that checks for that action and then executes the code that's written here. Remember to add the new middleware function to the giant list in the file middleware/all_middleware.

The code should look something like

```dart
reset_local_storage_middleware(Store<AppState> store, dynamic action, NextDispatcher next) {
  next(action);
  if (action is actions.ResetLocalStorage) {
    // code you wrote above
  }
}
```

dave-doty (changes requested):

lib/src/actions/actions.dart
```diff
@@ -238,6 +238,21 @@ abstract class LocalStorageDesignChoiceSet
   static Serializer<LocalStorageDesignChoiceSet> get serializer => _$localStorageDesignChoiceSetSerializer;
 }
 
+/////////////////////////////////////////////////////////////////////////////////////////////////////////////
+// set options for when to write design to localStorage
```
dave-doty, line 242: This comment seems like it was copied and pasted from somewhere else.

dave-doty (approved):

Issue #722, opened 2022-01-13
moving domain should be disallowed if overlapping another domain


damienwoods:
There seems to be a bug that allows me to make an illegal configuration, with overlapping strands. Example: 

Selected some domains, for dragging:
![Screenshot 2022-01-13 at 20 45 36](https://user-images.githubusercontent.com/40836180/149406752-c1a34ab2-c6b3-4716-ac01-c97c3b4f4547.jpg)

After dragging two strands are overlapping: 
![Screenshot 2022-01-13 at 20 46 08](https://user-images.githubusercontent.com/40836180/149407065-3beba0ac-6b8b-473a-9873-78270a5d9445.jpg)

Strands should have turned grey and system disallowed me from dragging to this position.

dave-doty:
There's a check for this, but there's some bug. Here's a minimal reproducible example (if there's only two domains in the design then the check succeeds and the domains cannot be moved to overlap):

![image](https://user-images.githubusercontent.com/19274365/149506740-2c6000a2-daee-44a6-9ff1-71d612e45616.png)

Moving the first domain over the second succeeds but should not:

![image](https://user-images.githubusercontent.com/19274365/149506919-450f0e27-109c-440a-86ba-a5412d1db1ed.png)


![image](https://user-images.githubusercontent.com/19274365/149506815-7699e1d4-8c22-48b6-adb6-5f3dda81d5c8.png)

However, if the third domain is not present, then the move is prevented as it should be:

![image](https://user-images.githubusercontent.com/19274365/149507025-b93bf9dd-219e-4eea-9979-f8c1cb63e4e3.png)

Issue #725, opened 2022-01-14
moved domains should be reselected after move


dave-doty:
Select these three domains:

![image](https://user-images.githubusercontent.com/19274365/149507118-ef2e8c0b-830e-43ec-8754-54ac61fabeb2.png)

And move them. They should still be selected after the move (in case the user wants to move them again), but they are not:

![image](https://user-images.githubusercontent.com/19274365/149507198-35ad17a2-a62f-436f-a782-c9d8d7eb3b44.png)

dave-doty:
This works properly for moving strands, and it is handled in middleware/reselect_moved_copied_strands.dart. I must have simply forgotten to include similar logic when implementing issue #416. 

It's non-trivial to implement because the domain objects after moving are different objects than the ones before the move. Look at the logic of middleware/reselect_moved_copied_strands.dart to see how we do it for strands; essentially recording addresses of strands that can be looked up after the reducer is applied (through the call to `next(action);` in the middle of the middleware function). This is one of those rare situations where it makes sense to do middleware logic both before and after the call to `next(action);`.

PR #726, opened 2022-01-14
#722 moving domain should be disallowed if overlapping another domain

dave-doty:


dave-doty:
Oops, this was a bad branch name. I accidentally copied and pasted the incorrect issue. But it is issue #722 being fixed in this PR, not issue #725.

Issue #681, opened 2021-11-19
merging strands should keep at least one existing name


dave-doty:
If two strands are merged by adding a crossover or ligating, then at least one of the two strand names should be preserved. If exactly one strand has a name, that should be the name of the new strand.

However currently if exactly one strand has a name, it is erased when merging the strands.

dave-doty:
More generally, if a property is nullable/optional (e.g., name), then if one strand has the property and the other doesn't, then assign the property from the strand that had it as non-null.

If both strands have the property, then take the new property value from the strand that was clicked (in the case of ligate) or clicked first (in the case of creating a crossover). This means `join_two_strands_with_substrands` should be re-written to take a Boolean as input indicating whether to take the new properties from `strand1` or from `strand2`.

dave-doty:
Also re-write nick reducer to take properties from original strand (e.g., new names are same as name of strand before).

PR #741, opened 2022-01-27
#681 merging strands should keep at least one existing name

cehsour:


dave-doty (approved):
Approved assuming unit tests pass; I had to resolve conflicts with dev.

Issue #731, opened 2022-01-18
show loading dialog when loading design


dave-doty:
When loading a large design, there is no indication that loading is being done. Show a dialog to the user that says "Loading design..." that goes away when the new design is loaded.

This is especially useful when loading large designs that take a while to load, but look visually similar to the previous design, so there's no obvious change when the design is finally loaded.

PR #755, opened 2022-03-04
Show loading dialog when loading design #731

sarahayu:


lib/src/middleware/load_file.dart
```diff
@@ -1,16 +1,29 @@
...
+import 'package:scadnano/src/actions/actions.dart';
 
 import '../actions/actions.dart' as actions;
 import '../app.dart';
 import '../state/app_state.dart';
 import '../util.dart' as util;
 
 load_file_middleware(Store<AppState> store, action, NextDispatcher next) {
-  next(action);
+  if (action is actions.PrepareToLoadDNAFile && !action.unit_testing) {
```
dave-doty, line 13: Add some comments here acknowledging that this is hacky and that we don't know why it works. Also enumerate the things we tried that didn't work. In the future I'm likely to get the idea that one of those ideas should work, and I'd like to be reminded that we already tried it.

dave-doty (changes requested):

dave-doty (approved):

PR #758, opened 2022-03-21
fixes #750: move submenus up if they go off the bottom of the page

### LINKED ISSUE #750: move submenus up if they go off the bottom of the page
Click on Help&rarr;Other versions. On my laptop, nothing below v0.15.1 is visible:

image

The submenu should move up to show as much as possible. However, it should not move things off the top of the page. That's especially for this menu, where the choices a user is most likely to want are at the top.

It would also be good to think about an option to scroll, or display the submenu in multiple columns, or some other way of displaying all items if possible. If that's easy to handle, it can be handled in the same branch that handles this issue; otherwise we can make it a separate issue.

sarahayu:


sarahayu:
note: this fix also adds scrolling functionality if moving the submenu upwards as far as it can is not enough to show the entire submenu

dave-doty (approved):

Issue #740, opened 2022-01-25
copy image of selected objects


dave-doty:
Currently one can copy strands or other objects. This puts a JSON representation of them in the system clipboard that can be used to paste into a text file, and if pasted into scadnano, it is interpreted to mean paste the objects somewhere in the scadnano design.

It is also possible to export SVG, but this exports the entire main or side view as a giant SVG object.

Make a new copy option (with a keyboard shortcut such as Ctrl+I) that reads the selected objects and copies to the system clipboard an image of them that can be pasted into programs such as slide presentation software. This might be done through SVG? Not sure. One should be able to select a strand, press Ctrl+I, then open Powerpoint and press Ctrl+V, and have an image of the strand be pasted into Powerpoint.

**UPDATE:** See comments on PR #769 for some information for why this feature is so difficult to implement. Seems to be browser security restrictions prevent putting certain types of data (e.g., SVG images) on the system clipboard.

dave-doty:
This might require hackery to find the SVG components that need to be exported, based on string `id` lookup in the rendered HTML. (i.e., reaching outside the React abstraction)

Maybe check out code for rectangle select for ideas.

dave-doty:
# Release notes

## Copy bitmap image of selected strands to clipboard
You can select some strands:

![image](https://github.com/UC-Davis-molecular-computing/scadnano/assets/19274365/8bf15c63-3fbf-4e95-b32c-1bdb0686543c)

and copy a (PNG bitmap) image of them to the clipboard by pressing Ctrl+I (or *Edit-->Copy image of selected strands to clipboard*), which can be pasted into other image programs such as Powerpoint or Inkscape:

![image](https://github.com/UC-Davis-molecular-computing/scadnano/assets/19274365/832a4701-a5d2-4eca-953b-8b56c96bf45c)

However, since this is a bitmap image, it will be pixelated on zoom-in:

![image](https://github.com/UC-Davis-molecular-computing/scadnano/assets/19274365/a075b739-7d5d-4c8c-b7e4-2b9d39fcdaad)

To save a vector graphics image (which will not pixelate on zoom-in), use *Export-->SVG of selected strands*. This is a bit more effort since it saves to a file, which then needs to be imported into Powerpoint/Inkscape/other image programs, but it will retain sharpness on zoom-in.

PR #769, opened 2022-04-29
Copy image of selected objects #740

sarahayu:


dave-doty (changes requested):
Hmm, I'm not sure exactly how this is implemented, but it isn't doing quite what I want. It is possible somehow to move SVG objects with the copy buffer. For example, you can copy SVG from one instance of Inkscape and paste into another instance of Inkscape, or even into Powerpoint (and vice versa), and it preserves all the SVG structure (groups and so forth). 

This implementation just copies a non-vector graphics image. I can copy from scadnano and paste into Inkscape or Powerpoint, but it does not preserve the SVG structure. I cannot copy several strands and then ungroup them to move them around. If I zoom in on Inkscape, the image is pixelated:

image


And if I look at the SVG source, it's an `img` object, not the original SVG:

image

It should be possible to copy the SVG directly. For instance, if I make these 16 squares in Inkscape and group them:

image

I can copy/paste it from Inkscape into Powerpoint, where I can then ungroup them (after responding "yes" to Powerpoint's request for me to convert the image into a drawing object):

image

Hopefully we should be able to use the same technique (whatever it is that lets you copy/paste between Inkscape and Powerpoint and preserve the structured vector graphics).

sarahayu:
It appears Chrome's clipboard and SVGs don't work well together (apparently for security reasons): [https://stackoverflow.com/a/31287434](https://stackoverflow.com/a/31287434)
The answer links to a Chromium bug report, the [latest comment ](https://bugs.chromium.org/p/chromium/issues/detail?id=504700#c9) by someone trying to do the same as us in copying an SVG to the clipboard. They said it has worked for them in Firefox, but I've tried changing our code to set the mime-type to image/svg+xml and tested it in Firefox to no avail.
We might just have to put this issue on standby until a workaround can be found?

dave-doty:
> It appears Chrome's clipboard and SVGs don't work well together (apparently for security reasons): https://stackoverflow.com/a/31287434 The answer links to a Chromium bug report, the [latest comment ](https://bugs.chromium.org/p/chromium/issues/detail?id=504700#c9) by someone trying to do the same as us in copying an SVG to the clipboard. They said it has worked for them in Firefox, but I've tried changing our code to set the mime-type to image/svg+xml and tested it in Firefox to no avail. We might just have to put this issue on standby until a workaround can be found?

Yeah, this is frustrating. I've searched google for any hint as to what exactly Inkscape and Powerpoint do to modify the copy buffer so that you can copy/paste SVG objects between them. Still no idea. Obviously it's there in the source code of Inkscape somewhere, but I don't know where to look. It's definitely not text. If you copy an Inkscape object and paste into a text editor, nothing happens.

abhishekgokhale:
Resolved conflict and formatted old code from this branch, branch should be good to be merged into dev with Ctrl/Cmd + I giving the PNG image on the clipboard

dave-doty (approved):

PR #772, opened 2022-06-21
fix #368 check offsets after reflecting strands to ensure offsets are…

### LINKED ISSUE #368: check offsets after reflecting strands to ensure offsets are within helix boundaries
On this design:

image

If you select reflect the strand horizontally, it goes outside the bounds of helix 1:

image

This should be checked and prevented in `check_mirror_strands_legal_middleware`.

sarahayu:
I don't know if test cases were necessary, but I added them. Though there are two problems with this:
1) Since the middleware check_reflect_strands_legal_middleware creates a window alert on error, dispatching the action StrandsReflect on an invalid mirror stalls the testing script until pressing "ok" in the browser.
2) This warning keeps popping up, which clogs up the testing output. 
```bash
04:24 +250: test/strand_mirror_unit_test.dart: StrandReflectValid strand_reflect_vertically_polarity_reverse
WARNING: some Action changed the design, so I am writing the Design to localStorage,
but that action is not UndoableAction, LoadDNAFile, Undo, or Redo
action is StrandsReflect {
  strands=[Strand {
    substrands=[Domain {
      helix=0,
      forward=true,
      start=0,
      end=10,
      deletions=[],
      insertions=[],
      is_first=true,
      is_last=false,
      is_scaffold=false,
      strand_id=strand-H0-0-forward,
      unused_fields={},
    }, Domain {
      helix=1,
      forward=false,
      start=5,
      end=10,
      deletions=[],
      insertions=[],
      is_first=false,
      is_last=true,
      is_scaffold=false,
      strand_id=strand-H0-0-forward,
      unused_fields={},
    }],
    is_scaffold=false,
    circular=false,
    modifications_int={},
    color=r: 247, g: 67, b: 8,
    unused_fields={},
  }],
  horizontal=false,
  reverse_polarity=true,
}
```
The bug seems to be fixed, and running the test script passes all test cases, aside from having to spam "ok" and the warnings that appear.
I could not find a similar test case in which a window alert is involved, so I'm not sure how to go around this. And I'm quite lost on why the warnings appear.

### DIFF
--- a/lib/src/middleware/check_mirror_strands_legal.dart
+++ b/lib/src/middleware/check_mirror_strands_legal.dart
@@ -35,6 +35,7 @@ check_reflect_strands_legal_middleware(Store<AppState> store, action, NextDispat
     var altered_design = design.remove_strands(strands_to_reflect);
     altered_design = altered_design.add_strands(reflected_strands);
 
+    // check overlapping strands
     try {
       altered_design.check_strands_overlap_legally();
     } on IllegalDesignError catch (e) {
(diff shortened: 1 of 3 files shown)


sarahayu:
I realize I had to cancel GitHub workflow since the test cases would have hanged

dave-doty:
Sorry, re-reading this issue, I want different behavior than originally specified in the issue. Check the part after **UPDATE**: #368

dave-doty (changes requested):
See previous comment

Issue #768, opened 2022-04-18
Rope select does not select correct points on switching to new design


sarahayu:
Rope select doesn't select correct points when switching to a new design. 
![image](https://user-images.githubusercontent.com/46186848/163890455-3cea65f4-cdad-45bd-a62c-436efdace385.png)
Steps to reproduce:
1. Load example 2_staple_2_helix_origami_deletions_insertions_mods
2. Load example 6_helix_origami_rectangle
3. Try rope select.

What's more, if between steps 1 and 2 you do a rope select, i.e.
1. Load example 2_staple_2_helix_origami_deletions_insertions_mods
2. Select 3 points using rope select (you don't even have to select any strand)
3. Load example 6_helix_origami_rectangle

The design does not autofit properly, and doing "autofit current design" does not even autofit UNLESS you move the design using the cursor, and somehow the bounding box reloads and then it's able to autofit the current design properly.
![image](https://user-images.githubusercontent.com/46186848/163891093-9ecee3a7-2631-4155-85e2-a055fca35ac9.png)

sarahayu:
The second bug mentioned with the autofit appears to not be exclusive to rope select. Step 2 wherein you rope select can be replaced with normal pointer select and the bug will behave the same way.

dave-doty:
I can't seem to reproduce this.

When you say "normal pointer select", does that mean clicking on a single strand? What about the selection box (i.e., dragging while pressing Shift in Select mode)?

sarahayu:
Yes, it means clicking on a single strand. Selection box does not seem to reproduce it. 
I've found it sometimes fails to reproduce the second bug at a few times, I think when you accidentally click some random spot in the design after doing a rope select (i.e. select a region, then clicking the design (without holding the Ctrl button) somewhere else). It then autofits properly in this case, but the rope select behavior is still wrong.

PR #775, opened 2022-06-26
#768 rope select does not select correct points on switching to new design

sarahayu:
Design will still not autofit, but rope select and pencil works as expected.


dave-doty (changes requested):
Can you add a comment explaining why we are calling `getSizes().realZoom` here instead of `getZoom()`? i.e., how was `getZoom()` causing the bug?

PR #782, opened 2022-07-01
closes #771 warn on unpaired deletion/insertion

### LINKED ISSUE #771: warn on unpaired deletion/insertion
Unlike cadnano, which associates deletions and insertions to helix positions rather than to domains, since scadnano associates them to individual domains, it allows deletions/insertions to be "unpaired", i.e., to have a deletion or insertion on one strand that is not on the strand to which it is bound:

image

This is allowed because it can arise during temporary Designs, for instance when pasting or moving a strand with a deletion next to another one that (initially) does not have a matching deletion.

However, it is intended for all insertions and deletions to be "paired": if there are two domains at the same helix and offset in opposite directions, then each should have an insertion/deletion if and only if the other one does, and if insertions, both should be the same number of bases.

By default, scadnano will display a warning symbol, the same one displayed on base or domain mismatches:

image

Whether such a warning is displayed can be configured through the View menu.

EdwinChang24:


lib/src/state/design.dart
```diff
@@ -506,6 +506,22 @@ abstract class Design with UnusedFields implements Built<Design, DesignBuilder>,
...
+  @memoized
+  BuiltMap<Domain, BuiltList<Mismatch>> get insertion_deletion_mismatches_map {
+    var insertion_deletion_mismatches_map_builder = MapBuilder<Domain, ListBuilder<Mismatch>>();
+    for (Strand strand in this.strands) {
+      for (Domain domain in strand.domains) {
+        insertion_deletion_mismatches_map_builder[domain] =
+            this._find_mismatches_on_substrand(domain, only_insertion_deletion: true);
+      }
+    }
+    var insertion_deletion_mismatches_builtmap_builder = MapBuilder<Domain, BuiltList<Mismatch>>();
```
dave-doty, line 518: This line seems strange to me. It doesn't look like there's any need to call `build()` on the `MapBuilder` named `insertion_deletion_mismatches_map_builder`.

Actually, in general it's good to imitate existing code (and maybe some nearby code looks like this), but in this case I'd do this differently. In particular, the built library provides [extension methods](https://dart.dev/guides/language/extension-methods#:~:text=Extension%20methods%2C%20introduced%20in%20Dart,extension%20methods%20alongside%20regular%20methods.) so that you can call `build()` directly on native collections such as `List`'s and `Map`'s. I find that more readable. So I'd implement this something like (haven't tested this code though):

```dart
var insertion_deletion_mismatches_map = Map<Domain, List<Mismatch>>();
for (Strand strand in this.strands) {
  for (Domain domain in strand.domains) {
    insertion_deletion_mismatches_map[domain] =
        this._find_mismatches_on_substrand(domain, only_insertion_deletion: true);
  }
}
var insertion_deletion_mismatches_half_built_map = Map<Domain, BuiltList<Mismatch>>();
insertion_deletion_mismatches_map.forEach((domain, mismatches)) {
  insertion_deletion_mismatches_half_built_map[domain] = mismatches.build();
}
return insertion_deletion_mismatches_half_built_map.build();
```

Which is a bit cleaner to read, and doesn't do the unnecessary `build()` on the first data structure.

lib/src/state/design.dart
```diff
@@ -1464,7 +1480,7 @@ abstract class Design with UnusedFields implements Built<Design, DesignBuilder>,
     }
   }
 
-  ListBuilder<Mismatch> _find_mismatches_on_substrand(Domain substrand) {
+  ListBuilder<Mismatch> _find_mismatches_on_substrand(Domain substrand, {bool only_insertion_deletion = false}) {
```
dave-doty, line 1483: I think I'm a bit confused what's happening here. The class `Mismatch` already exists, and so does this private method `_find_mismatches_on_substrand`, is that correct? If I recall `Mismatch` specifically refers to a mismatching DNA base pair, e.g., an A next to a G.

A new class should be created that specifically represents an unpaired deletion or insertion. In fact, please rename the actions and field names above to use the phrase "unpaired" instead of "mismatch". The concepts are slightly different so it's good to use a different word. For example, a DNA base on one strand does not mismatch the adjacent strand if the adjacent strand has no DNA sequence assigned, but this would be "unpaired" if we were talking about a deletion.

More importantly, the class `Mismatch` was designed specifically for DNA bases: it has a field `int within_insertion` that refers to the mismatching base being within an insertion in the domain, whereas no unpaired deletion could be "within" an insertion. Since the `Mismatch` class doesn't model the data we want to describe for unpaired deletions/insertions, it should not be used. Let's make a brand new class.

It might be a good idea to rename `Mismatch` to `DNABasePairMismatch` to make it clear that this is the only thing it refers to.

Now that I think about it, we don't even need a new state class to represent unpaired insertions/deletions. The class `Address` already captures exactly what we want. In other words I think the memoized getter above `insertion_deletion_mismatches_map` should, first, be renamed `unpaired_insertion_deletion_map`, then have its type be `BuiltMap<Domain, BuiltList<Address>>`. In other words each `Domain` just maintains a list of `Address`'es of unpaired insertions/deletions. The view code can use this to figure out where to put the red stars.

This might require a bit of refactoring to create a common View component that can be used for all three purposes: DNA base mismatches, Domain name mismatches, and unpaired insertions/deletions.

dave-doty (changes requested):
See comments above; we can discuss in person if necessary.

dave-doty (changes requested):
Sorry, can I be pedantic and request that we rename things (actions, functions, etc.) to refer to *unpaired* insertions and deletions? (rather than "mismatches", which is a different concept)

dave-doty (approved):

dave-doty:
Sorry, I forgot about something else I had wanted implemented. I think I said it in a meeting but didn't write it down.

Let's make a new menu View-->Warnings, and move the following there:

- Show DNA base mismatches
- Show unpaired insertions/deletions
- Show domain name mismatches

dave-doty:
Okay, I implemented the View menu re-arrangements in the dev branch.

One thing that's odd is that the warnings stars come in pairs for insertions but not deletions:

image

Can we change the insertions to be like deletions, where it only shows the warning on the strand lacking the insertion?

sarahayu:
We have the warning show on insertions that are of different lengths, so that's why it appears on both strands. Should the warning appear only on one of them in the case of two insertions with different lengths?

dave-doty:
> We have the warning show on insertions that are of different lengths, so that's why it appears on both strands. Should the warning appear only on one of them in the case of two insertions with different lengths?

Ah, I see. Keep the warning on both if they are different lengths.

But is it possible to slide it out a bit in this case, so it's more visible? It shows up underneath the insertion so is hard to see.

sarahayu:
> But is it possible to slide it out a bit in this case, so it's more visible?

Ok, should it still be slid out if there's no insertion?

dave-doty:
No it's only necessary to keep it from being hidden by the insertion.

Issue #36, opened 2019-12-14
allow helix index changing in batch


dave-doty:
Allow both the index of each helix in the list of all helices to be altered, and also the view order specified in `DNADesign.helices_view_order`.

Furthermore, now that helices are not constrained to be consecutive (#172), allow to change helix indices to be any integers.

dave-doty:
This was partially addressed by commit 2eab946, but I'm leaving it open to allow a faster way to re-index all helices at once. Currently they can only be changed one at a time. Swapping two is particularly annoying since one needs to assign one of them a temporary index not belonging to any helix to avoid collision.

dave-doty:
Unlike `helices_view_order`, which has a meaning that does not depend on the old `helices_view_order`, re-numbering helices requires that you know the old helix numbers when assigning new ones. So entering it the same way as `helices_view_order` would be harder to use, since, after erasing the old list of indices, one would have to remember what order they appeared in for the new indices to make sense.

So, it seems best to implement this in a more structured way, where the hold indices are displayed read-only, with each of them next to a text field where the new index can be chosen.

PR #784, opened 2022-07-22
36 allow helix index changing in batch

vbKenobi:
## Description

Added a view component called adjust helix indices under the group dropdown, that allows user to change helix indices. 

## Related Issue
#36-allow_helix-index-changing-in-batch

## Motivation and Context
Allows user to change the group indices under the group --> adjust helix indices menu dropdown option. 

## How Has This Been Tested?
Manually used the menu option to make sure that helix indices would change.


dave-doty:
Can you fill in a bit under each heading of the PR?

vbKenobi:
Should be updated!

dave-doty (changes requested):

lib/src/middleware/helix_idxs_change.dart
```diff
@@ -12,6 +12,24 @@ helix_idxs_change_middleware(Store<AppState> store, dynamic action, NextDispatch
...
+    var new_indices = action.idx_replacements.values.toSet();
+
+    if (new_indices.length != action.idx_replacements.length) {
+      Set<int> intsSoFar = {};
+      List<int> all_helices = action.idx_replacements.values.toList();
+
+      for (var old_index in old_idxs) {
+        var new_index = action.idx_replacements[old_index];
+        if (intsSoFar.contains(new_index)) {
+          var msg = 'First duplicate found for helix ${old_index} with new index ${new_index}. ';
```
dave-doty, line 24: Sorry, try it one more time:

image

Imagine you're a user and there's a HUGE list of indices there. If they tried to assign index a to b and also index c to b, then they'll want to know all three of a, b, c. The current error message is only telling them two of these.

Try this error message: "You tried to assign helices 4 and 5 both to 0. Each helix must have a *unique* new index; make sure all the integers you write are distinct from each other and do not appear elsewhere in the design."

dave-doty (changes requested):

dave-doty (changes requested):
Something is still not right. I try this invalid change:

image

It gives this error message:

image

It should instead say "*You tried assigning existing helices 0 and 1 to 0*" In other words it should indicate both "source" indices that are in conflict. Otherwise the user cannot locate the conflict and fix it if they only know one of the helices causing the problem.

dave-doty (approved):

PR #786, opened 2022-07-27
closes #777: update project configuration

### LINKED ISSUE #777: add dart format configuration options to text file stored in repo
This will ensure that if we run dart format, we all run it the same way.

EdwinChang24:
## Description

* Adds a `.editorconfig` file with some universal configuration options. See more about EditorConfig [here](https://editorconfig.org/).
* Updates `.gitignore` and `.gitattributes` to use templates.
* Adds configuration for Visual Studio Code and JetBrains IDEs, with the 110-character max line length configured.

## Related Issue

Resolves the problems mentioned in #777.

## Warning

If you are using a JetBrains IDE, you may have to force pull when pulling these changes to override the local files. I haven't tested it much.


dave-doty (commented):
I guess I'm a bit reluctant to put all these IDE config files in the repo. I know from experience with using multiple computers that sometimes the IDE config files contain absolute pathnames, which breaks between different computers with different directory structures.

This EditorConfig sounds great, but I wonder if there's a way to use it where we only specify one single text file that it recognizes (and that Intellij and Visual Studio Code recognize), but where we don't need to put all the other IDE config files in the repo.

dave-doty:
We did this more elegantly through #864

Issue #34, opened 2019-12-13
support Extensions on the end of a Strand


dave-doty:
This would give a nice way to specify toeholds and extensions common in DNA strand displacement and structural designs.

See also https://github.com/UC-Davis-molecular-computing/scadnano-python-package/issues/2.

# Display

One thing that is not straightforward is how to display such extensions. One idea is to make it a straight line at an angle, with length proportional to the number of DNA bases. Some complications:

1. What angle? 

2. Should the angle be customizable? 

3. If so, independently for each extension, or just one option in the View menu to apply to all of them?

4. If it is customizable individually for each extension, then the natural place to store it is in the Design json, rather than as an AppUIState field. But this is a bit ugly; there's very little other UI specifications in the Design object.

5. Should the length be equivalent to a Domain of the same length? Or should it be shorter, with the DNA bases rendered closer together, to reduce the potential for visual clutter?

# Workarounds
If you want to put a single-stranded extension/overhang on a strand before this feature is implemented, there are two solutions. The first is recommended:

1. **5' or 3' modification:** Add a 5' or 3' modification to the strand, and specify its `idt_text` to be the sequence you want. Normally the idt_text is a code for a DNA modification, e.g., `/5Biosg/` for 5' biotin. However, it is simply placed inline with the DNA sequence when creating an IDT formatted file. So you can simply give a DNA sequence. For example, here is a screenshot showing how to add a tail of 6 T's to the 3' end of a strand:

    ![image](https://user-images.githubusercontent.com/19274365/99155481-1a54fc00-266d-11eb-9aaa-8d5a10236cbf.png)

2. **add crossover to a "fake" helix:** This is less elegant, but you can also create a "fake" helix where the single-stranded part is represented as a `Domain` instead of an `Extension`:

    ![image](https://user-images.githubusercontent.com/19274365/99155446-df52c880-266c-11eb-8c33-faed0dcafd35.png)

dave-doty:
I removed the "High Priority" label, since the workaround with 5'/3' modifications is fairly simple and actually not that hacky.

dave-doty:
Note: after some consideration, it makes sense to have this be a third type of Substrand called Extension. Also, the position of the end of it (the one not connected to the adjacent Domain) should be customizable by the user, by dragging in the web interface, or using units of nanometers for the Python interface.

dave-doty:
[...]

PR #805, opened 2022-09-10
789 saved dialog for future

homaho1:


dave-doty (approved):
I haven't had much time to review this. Can you please make sure you test it out a bit with the features mentioned here before merging? https://github.com/UC-Davis-molecular-computing/scadnano/issues/34#issuecomment-1241023201

lib/src/view/design_main_strand.dart
```diff
@@ -552,7 +552,12 @@ after:
     items[num_bases_idx] = DialogInteger(
         label: 'number of bases', value: 5, tooltip: 'number of bases to include in this extension');
 
-    var dialog = Dialog(title: 'add extension', items: items, type: DialogType.add_extension);
+    var dialog = Dialog(title: 'add extension', items: items, type: DialogType.add_extension, process_saved_response: (saved_items) {
```
dave-doty, line 555: Actually, can we re-do this to keep the original behavior? The original behavior was actually to save nothing from the previous dialog. There's also logic that only gives an option to add an extension where there isn't currently one. For example, if there's a 5' extension and no 3' extension, the original behavior was for there only to be one option for where to add the extension, 3'.

For this it might be simplest to add an optional keyword argument to the Dialog constructor that simply tells it not to use any saved responses when first rendering the dialog, i.e., just go with what the original Dialog object says.

dave-doty (changes requested):

dave-doty (approved):

PR #808, opened 2022-09-17
closes #807: Add option to disable PNG caching for DNA sequences

### LINKED ISSUE #807: Option to disable PNG caching of DNA sequences
Lots of displayed DNA sequences slow down the panning and zooming when displayed as SVG. So when zoomed far enough out scadnano creates a PNG to display instead which is much faster to render.

Give a View option to disable this, which will help with debugging problems related to it. (e.g., see #806)

EdwinChang24:


lib/src/util.dart
```diff
@@ -1497,10 +1497,11 @@ void svg_to_png_data() {
 /// and the zoom is not above threshold `is_zoom_above_threshold`,
 /// and there is no pending action `disable_png_cache_until_action_completes`.
 bool use_png(String dna_sequence_png_uri, bool is_zoom_above_threshold,
-    actions.Action disable_png_cache_until_action_completes) {
+    actions.Action disable_png_cache_until_action_completes, bool disable_png_caching_dna_sequences) {
   return dna_sequence_png_uri != null &&
       !is_zoom_above_threshold &&
-      disable_png_cache_until_action_completes == null;
+      disable_png_cache_until_action_completes == null &&
```
dave-doty, line 1503: This isn't a problem with this PR, it seems it was already present. But I can't tell what this field `disable_png_cache_until_action_completes` does. I can't find anywhere in the codebase where it gets set. Is it ever non-null? If not then it is probably just a remnant of some earlier implementation and should be removed.

dave-doty (changes requested):

dave-doty (approved):
Approved subject to testing with latest commit.

dave-doty (approved):

Issue #745, opened 2022-02-02
modifications should be copied with strand


dave-doty:
Create a strand with a modification:

![image](https://user-images.githubusercontent.com/19274365/152145201-432914f6-bd46-46c7-ba8f-98841b98015b.png)

Copy and paste it. The pasted strand lacks the modification, but should have it:

![image](https://user-images.githubusercontent.com/19274365/152145221-47f44faf-186b-482e-a65b-2d649853da22.png)

dave-doty:
This is somewhat tricky when copying from one design to another. This is because modification data is not stored directly in the strand JSON, only the `id` field that is used as a key to access the full Modification data elsewhere in the Design JSON.

For example, here is what is copied to the clipboard after selecting the first strand above:

```json
{
  "strands": [
    {
      "color": "#32b86c",
      "domains": [
        {"helix": 6, "forward": true, "start": 10, "end": 31}
      ],
      "5prime_modification": "5'-Q"
    }
  ],
  "helices_view_order": [0, 1, 2, 3, 4, 5, 6, 7, 8, 9, 10]
}
```

In the Dart object, the full modification *is* stored in the Strand object, in this case as `Strand.modification_5p`, but only the `id` is written when serializing to JSON:

```dart
if (this.modification_5p != null) {
  json_map[constants.modification_5p_key] = this.modification_5p.id;
}
```

Probably the best way to handle this is the following. When copying Strand data to the clipboard, store any modifications as a separate top-level field in the JSON, under the key "modifications_in_design" (just like when writing a `.sc` file). Only store each modification once, just like with that scheme, i.e., if 2 different strands both have a modification with the same `id`, only store one copy of the modification. The storage should be just like the `.sc` file, for instance

```json
{
  "strands": [
    {
      "color": "#32b86c",
      "domains": [
        {"helix": 6, "forward": true, "start": 10, "end": 31}
      ],
      "5prime_modification": "5'-Q"
    }
  ],
  "helices_view_order": [0, 1, 2, 3, 4, 5, 6, 7, 8, 9, 10],
  "modifications_in_design": {
    "5'-Q": {
      "display_text": "Q",
      "idt_text": "Q",
      "display_connector": true,
      "location": "5'"
    }
  }
}
```

dave-doty:
Closed by #823

PR #820, opened 2022-10-07
#745 modifications should be copied with strand

sarahayu:


lib/src/middleware/system_clipboard.dart
```diff
@@ -127,13 +130,44 @@ void put_strand_info_on_clipboard(Store<AppState> store) {
       helices_view_order = group.helices_view_order.toList();
     }
 
+    //Get modifcations from all strands
+    var modifications = all_modifications(strands.toBuiltList());
```
dave-doty, line 134: Any particular reason the `BuiltSet<Strand>` is being converted to a `BuiltList<Strand>` here? If there's a specific reason, comment it, otherwise you could just change the type of the parameter of `all_modifications` to be `Iterable<Strand>` or `BuiltSet<Strand>`.

lib/src/reducers/strands_copy_info_reducer.dart
```diff
@@ -101,7 +102,23 @@ Tuple2<List<Strand>, List<int>> parse_strands_and_helices_view_order_from_clipbo
...
-
+  
+  Map mod_jsons = clipboard_json[constants.design_modifications_key][0];
+  Map<String, Modification> mods = {};
+  for (var mod_key in mod_jsons.keys) {
+    Modification mod;
+    try {
+      mod = Modification.from_json(mod_jsons[mod_key]);
+    } on Exception {
+      print(error_msg);
```
dave-doty, line 113: Hmm, here and elsewhere in this function, I find it odd that we print the error message and return null. Is this handled gracefully for this function to return null? If there's an error here let's try to make sure it doesn't do something super strange.

dave-doty (approved):

dave-doty (changes requested):
See comments above.

lib/src/reducers/strands_copy_info_reducer.dart
sarahayu: It seems the function is under the impression it will return null on JSON decode error, so we've added a comment to let the user know to check for it when using this function. Is this alright?

dave-doty: Looks good.

dave-doty (approved):

PR #821, opened 2022-10-09
fixes #819 change color of warning star for domain name mismatch

### LINKED ISSUE #819: change color of warning star for domain name mismatch
It is too easy to confuse it with a DNA base mismatch; see issue #818.

This should be done by adding a prop `color` to `DesignMainWarningStar`, which is red for a DNA base mismatch but some other color for domain name mismatches. Perhaps blue or green, but it should be something that stands out in a large design and is easy to notice.

homaho1:


dave-doty (changes requested):
This doesn't seem like a very extensible solution to go through CSS for the color. Is it possible to set the property directly in the React component? Then for the future we could add more colors more easily without having to edit the CSS style.

homaho1:
I used `style` instead of `classname`. Is that the way you wanted it to be implemented?

dave-doty:
> I used `style` instead of `classname`. Is this the way you wanted it to be implemented?

Sure, as long as it's straightforward to use a new color if we want and only have to change that one React prop.

dave-doty (approved):

homaho1:
No problem. Should I go ahead and merge pull request?

dave-doty:
Yes

PR #822, opened 2022-10-10
fixes #817 many editable values in Dialog forms should be existing va…

### LINKED ISSUE #817: many editable values in Dialog forms should be existing value of strand/domain/etc., not value entered last time dialog was open
An exception to the new saved dialog feature: If you assign DNA to a strand that already has a DNA sequence assigned, the initial value should be the existing sequence, not the sequence that was there the last time the dialog was open.

homaho1:


dave-doty (approved):
Looks good, although I'm slightly confused about one behavior, see comment.

homaho1:
It doesn't. By setting `use_saved_response: false`, we disable auto-populating with the saved responses. It would only show the actual value it has.

PR #823, opened 2022-10-19
Dev

dave-doty:


dave-doty:
Closes #745: modifications should be copied with strand

Issue #825, opened 2022-10-26
Importing deletions on 5'/3' staple end from Cadnano


mmguar:
I am importing a cadnano file that has several deletions that correspond to 5'-3' ends of staples.
From previous issues reported, this does not seem to be allowed by scadnano, which wants you to just design the staple one base shorter.
This is what it looks like:
![image](https://user-images.githubusercontent.com/56086936/198115503-ed152df6-0855-42ee-8ffe-52f0aa26f756.png)
I assume the balloon is telling me that there is something wrong there? Assigning a sequence does not work either, until I change the length of the staple.

Thus I was wondering if it would be possible to modify the candnano import to automatically fix these, as it can be ab it tedious by hand.

Thanks!

dave-doty:
That's correct, in scadnano deletions and insertions can only appear "interior" to a domain, not on an end. So the import is ignoring the deletion on the 3' end, and since the other deletion on the scaffold is still present, it's warning you that there's an unpair deletion.

We could try to come up with something to fix in the import, but I won't be able to get to it soon unfortunately, so for now you'll have to do it by hand or with a Python script.

dave-doty:
Ideally we have a more explicit warning, where if the import wants to get rid of a deletion, it warns the user or offers them the option to shift the deletion. (the latter could be a bit tricky to cover all the cases)

dave-doty:
See also https://github.com/UC-Davis-molecular-computing/scadnano-python-package/issues/260

PR #826, opened 2022-10-27
Should fix 825: import cadnano -  domains that end on deletions

mmguar:
Import from cadnano can now shorten domains that end on a deletion (or multiple deletions).

## Description
Modified design.dart to look at domains and recognize which ones end on deletions.
Possibly unwanted behavior: it will shorten a scaffold too, if deletions are at either of its ends. 
It might not be the best way of doing this, and it's for sure not elegant, but it might be helpful to have a first implementation.

## Related Issue
https://github.com/UC-Davis-molecular-computing/scadnano/issues/825

## Motivation and Context
Avoids warning stars when importing from cadnano, automates a tedious compatibility task.

## How Has This Been Tested?
Tried on multiple deletions located at strand ends.
Could not run extensive tests on this machine (no google chrome).

### DIFF
--- a/lib/src/state/design.dart
+++ b/lib/src/state/design.dart
@@ -2125,6 +2125,23 @@ abstract class Design with UnusedFields implements Built<Design, DesignBuilder>,
         else
           start = old_base;
 
+	bool running = true;
+	while(running){
+	  switch(Design._cadnano_v2_import_extract_end_deletions(vstrands[old_helix]['skip'], start,end)){
+	    case -1:{
+	      start = start + 1;
+	    }
+	    break;
+	    case 1:{
+	      end = end-1;
+	    }
+	    break;
+	    case 0:{
+	      running = false;
+	    }
+	  }
+	}
+
         domains.add(Domain(
             is_scaffold: strand_type == 'scaf',
             helix: old_helix,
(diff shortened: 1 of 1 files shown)


dave-doty (changes requested):
I'm a bit worried that this doesn't handle corner cases well. For example, what if there is a strand with a domain of length 2 and a deletion on one of its ends? I think it will have an infinite loop moving the deletion back and forth between the ends:

image

I'm also not sure how to handle corner cases such as this:

image

I'm happy to say those cases are super-rare and just not handle them (i.e., remove the deletion), but definitely don't want things to break or crash in that case.

mmguar:
I haven't seen the issue in the second case yet, the top strand is shortened by one. Is it because either of them could be considered scaffold?

The first case does not hang in infinite loop, but still returns an error (I suppose when the staple is found not to exist anymore). 

```
****************************************************************************************************
* error type:    JSNoSuchMethodError
* error message: TypeError: can't access property Symbol.iterator, deserialized_helices_list is null
****************************************************************************************************
```
I am not sure what the best fix is. The error given is much clearer in the case where an entire staple has deletions. In that case the error given is Invalid Design.

Hopefully the code is still somewhat helpful in the future, I might have some time to improve on it, but not likely soon.

PR #828, opened 2022-11-08
closes #827: add "select all in current helix group" 

### LINKED ISSUE #827: add "select all in current helix group"
Currently, selecting all objects (Ctrl+A) selects all in the design. Add this to the Edit menu with the title "Select all".

Also add a new option "Select all in current helix group".

Strands spanning multiple helix groups should not be selected in the latter case.

EdwinChang24:


lib/src/actions/actions.dart
```diff
@@ -1409,6 +1409,18 @@ abstract class DeleteAllSelected
   String short_description() => "remove all selected items";
 }
 
+// Selects all selectable items in the current helix group.
```
dave-doty, line 1412: Instead of creating a new Action, I think it would be cleaner to add a Boolean field `current_helix_group_only` to the existing action `SelectAllSelectable`. Then the reducer `select_all_selectables_reducer` would essentially just contain the code currently in `select_all_selectables_helper` but would check that field instead of the extra parameter `only_in_helix_group`.

lib/src/view/menu.dart
```diff
@@ -386,6 +386,20 @@ with some default direction chosen. Play with it and see!
...
       ///////////////////////////////////////////////////////////////
+      // select all
+      DropdownDivider({}),
+      (MenuDropdownItem()
+        ..on_click = ((_) => window.dispatchEvent(new KeyEvent('keydown', keyCode: KeyCode.A, ctrlKey: true).wrapped))
+        ..display = 'Select All'
+        ..tooltip = '''\
+Select all strands in the design.'''
+        ..keyboard_shortcut = 'Ctrl+A')(),
+      (MenuDropdownItem()
```
dave-doty, line 397: Can we also make a keyboard shortcut Ctrl+Shift+A for this?

See view/design.dart around line 605 for an example of how to handle this:

```dart
// can paste even if nothing selected or not in select mode, if something is in copy buffer
if ((ev.ctrlKey || ev.metaKey) && key == KeyCode.V && !ev.shiftKey) {
  // Ctrl+V
  paste_strands_manually();
} else if ((ev.ctrlKey || ev.metaKey) && ev.shiftKey && key == KeyCode.V) {
  // Ctrl+Shift+V
  paste_strands_auto();
}
```

dave-doty (changes requested):

EdwinChang24:
Should we add test cases for this?

dave-doty (approved):

PR #846, opened 2023-03-04
845 click dragging extension ends crashes

homaho1:
Same issue as https://github.com/UC-Davis-molecular-computing/scadnano/pull/658


dave-doty (changes requested):
Can you add a comment at the relevant fixed part, and an explanation of why it is needed? I'm not quite seeing which change fixes the issue.

dave-doty (approved):

Issue #570, opened 2021-02-07
Export sequences to CanDo


UnHumbleBen:
The csv output format from scadnano is pretty close to that required by CanDo. The .csv from scadnano needs to include headings and colors of strands.

Related issue in [Python scripting library](https://github.com/UC-Davis-molecular-computing/scadnano-python-package/issues/167)

PR #850, opened 2023-04-04
CanDo Compatible DNA Sequence export and Dockerfile for running containerised. (Fixes #570)

headblockhead:
**Note: Sorry about this, I based my fork on the main branch and not the dev branch, so some commit history is from the main branch. Perhaps a squash merge to remove the extra commits?**

This pull request adds two files and makes changes in a few. 

It adds a `Dockerfile` ( A set of instructions that allow for reliable development on multiple platforms without changes to the host operating system ) and a `run-docker.sh` file to run the Dockerfile with correct arguments.

It changes the Export dropdown to add a new feature - 'CanDo DNA Sequences'. This exports the file in the cadnano2 format, accepted by CanDo.

## Description
I have added a `ExportCanDoDNA` class, alongside the `ExportDNA` class. This is used separate from the [Export > DNA sequences] dialogue as certain settings need to be set in order for CanDo to accept the CSV. ( Include Scaffold: false, Include only selected strands: false, Sort stands: true, Column Major order: true, Strand part to sort by: 5' ), so this would not work with the existing dialogue. I have added a case for `ExportCanDoDNA` in `middleware/export_dna_sequences.dart` that uses these settings, along with the main `cando_compatible_csv_export` function itself. This function is called by the `cando` ExportDNAFormat, which is hidden from the [Export > DNA sequences] dialogue by the fact that it is not assigned a name in the `_toString_map` of the `ExportDNAFormat` enum.

## Related Issue
This pull request should solve issue #570, requesting an export for CanDo feature since Feb 7, 2021.

## Motivation and Context
I wanted to add this feature as manually writing the CSV file and copying color values etc. is a lot of manual work when it could be done by scadnano easily.

## How Has This Been Tested?
I tested this by comparing it to manually edited `.csv` files that I confirmed worked with CanDo, and it created the exact same result. Unit tests may need to be written for this in the future and I have not had chance to run the existing tests on my changes locally, so hopefully GitHub Actions can confirm the tests for me.

## Screenshots (if appropriate):
image
image

Thanks for checking out my code! Sorry if I made any breaking changes or have not followed the design organisation of the project, but I have made my best effort to keep it simple.

PR #852, opened 2023-04-08
Add CanDo Atomic model CSV Export

### LINKED ISSUE #787: Export colors to CSV file
A useful feature of cadnano is the ability to export strands with their color saved in a column, which lets one group strands very easily for any sort of post-processing. I might have missed it, but I could not find a way to do it when exporting a scadnano csv.

headblockhead:
This pull request changes a few files. It is half of PR #850, is based on the dev branch and conversation about CanDo Export should be continued here.

It changes the Export drop-down to add a new feature - 'CanDo DNA Sequences'. This exports the file in the cadnano2 format, accepted by CanDo's Atomic Model generator.

## Description
I have added a `ExportCanDoDNA` class, alongside the `ExportDNA` class. This is used separate from the [Export > DNA sequences] dialogue as certain settings need to be set in order for CanDo to accept the CSV. ( Include Scaffold: false, Include only selected strands: false, Sort stands: true, Column Major order: true, Strand part to sort by: 5' ), so this would not work with the existing dialogue. I have added a case for `ExportCanDoDNA` in `middleware/export_dna_sequences.dart` that uses these settings, along with the main `cando_compatible_csv_export` function itself. This function is called by the `cando` ExportDNAFormat, which is hidden from the [Export > DNA sequences] dialogue.

## Related Issue
This pull request fixes issue #570, requesting an export for CanDo feature since Feb 7, 2021.

## Motivation and Context
I wanted to add this feature as manually writing the CSV file and copying color values etc. is a lot of manual work when it could be done by scadnano easily.

## How Has This Been Tested?
I tested this by comparing it to manually edited `.csv` files that I confirmed worked with CanDo, and it created the exact same result. Unit tests may need to be written for this in the future.

image
image

Thanks for checking out my code! Sorry if I made any breaking changes or have not followed the design organisation of the project, but I have made my best effort to keep it simple.

### DIFF
--- a/lib/src/actions/actions.dart
+++ b/lib/src/actions/actions.dart
@@ -1924,6 +1924,27 @@ abstract class ExportDNA with BuiltJsonSerializable implements Action, Built<Exp
 }
 
 /////////////////////////////////////////////////////////////////////////////////////////////////////////////
+// Export CanDo CSV DNA
+
+abstract class ExportCanDoDNA with BuiltJsonSerializable implements Action, Built<ExportCanDoDNA, ExportCanDoDNABuilder> {
+  ExportDNAFormat get export_dna_format;
+  factory ExportCanDoDNA(
+  {
+   ExportDNAFormat export_dna_format,
+  }) {
+    return ExportCanDoDNA.from((b) => b
+    ..export_dna_format = export_dna_format);
+  }
+
+  factory ExportCanDoDNA.from([void Function(ExportCanDoDNABuilder) updates]) = _$ExportCanDoDNA;
+
+  ExportCanDoDNA._();
+
+  static Serializer<ExportCanDoDNA> get serializer => _$exportCanDoDNASerializer;
+}
+
+
+/////////////////////////////////////////////////////////////////////////////////////////////////////////////
 // Export SVG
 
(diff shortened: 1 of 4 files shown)


dave-doty (approved):
Thanks, this works for me now even with Redux dev tools turned on.

One question: it sounds like this export format is supported by cadnano. It happens to be useful for importing into CanDo, but it isn't really a "CanDo" format. Perhaps it would be better to name the menu option "DNA sequences with color" or "DNA sequences (cadnano format)" something similar, and the tooltip can explain that it is useful for importing to CanDo. But I'm not really familiar with these file formats. If it's only used by CanDo, then perhaps it is appropriate to call it "CanDo". I was vaguely aware that CanDo does have its own file format (accepted by the command line tool but not the web interface), so I want to avoid confusion with that.

What do you think? I'm fine either way, but I want to make sure it is not confusing for other people.

headblockhead:
Thanks for merging! I think that sounds like a good idea. The format is the same as exported by cadnano v2. Perhaps a title of "cadnano v2 DNA sequences" with a tool-tip of "Export design's DNA sequences as a CSV in the same way as cadnano v2. This is used with CanDo's atomic model generator."

PR #853, opened 2023-04-08
Add Dockerfile for containerised running.

headblockhead:
This PR adds a Dockerfile and `run_docker.sh`. It is half of PR #850, is based on the dev branch and conversation about the Dockerfile should be continued here.

If you don't want to add this, feel free to close this - it is not a very important addition.

## Description

A `Dockerfile` is a file that allows for reliable running of programs and services by providing a list of commands to be run in an isolated (*ish*) environment. 

#### Docker Docs
https://docs.docker.com/get-started/overview/
https://docs.docker.com/engine/reference/builder/

#### Docker in 100 seconds
https://www.youtube.com/watch?v=Gjnup-PuquQ

## Motivation and Context

I want to add this to allow for easy running of scadnano on many platforms without having to install anything on the host computer.

## How Has This Been Tested?

I have tested the file on other computers, and it successfully runs on linux at port 8080. Changes made to the code will apply in the container automatically.


dave-doty (changes requested):
Do you mind adding a section to the README.md (or CONTRIBUTING.md if that's the main use case) with instructions for this? e.g., a link to Docker download/installation instructions, and how to use it the `run_docker.sh` file. (even if it's as simple as running `./run_docker.sh` at the command line) Also if it's only tested on e.g., Linux, maybe saying that.

Also, a short explanation of what it accomplishes (I assume from the last Dockerfile command, lets you run a local scadnano server without having to follow the setup instructions in the CONTRIBUTING document.

Thank you!

headblockhead:
Hello! I wrote some instructions - does that work?

dave-doty:
Looks great! I haven't tried it myself but assuming it works, this should be a great alternative way to install scadnano locally.

dave-doty (approved):
See comment.

PR #856, opened 2023-04-28
836 upgrade colorpicker

sarahayu:
I renamed `componentWillReceiveProps` to `UNSAFE_componentWillReceiveProps` as per the warning's recommendation. I could not find any updated distribution of react-color.min.js, so I modified the minified js file. 
(The most recent distribution I [found](https://unpkg.com/react-color@3.0.0-beta.3/dist/react-color.min.js) appears unchanged from the one we currently have, and still uses the deprecated `componentWillReceiveProps` method.)
Technically, renaming all instances of the deprecated function `componentWillReceiveProps` to anything else like `nonexistentfunction` will remove the warning, but the code that uses the deprecated function will fail silently. The only way I could tell this solution works is that one part of the code that uses `componentWillReceiveProps` is responsible for turning the hex code uppercase when the user loses focus on the hex color text input (input 'ffffff' into the hex color value, click somewhere else, and it will turn into 'FFFFFF'). If the function is not properly renamed, the input will remain lowercase and give no warnings/errors in console. Using `UNSAFE_componentWillReceiveProps` removes the warning AND preserves the function.


dave-doty (approved):
Thanks for the explanation, I would have had a hard time understanding the file difference. :)

PR #866, opened 2023-06-28
closes #849-give-more-useful-error-message-when-sc-file-format-is-incorrect

### LINKED ISSUE #849: Give more useful error message when `.sc` file format is incorrect
I encountered an error while reading the file Origami-hinge-ok.json:
but CadnanoSQ can open it which I received from my collegue.
image

****************************************************************************************************
* error type:    NullError
* error message: NoSuchMethodError: method not found: 'get$iterator' on null
****************************************************************************************************

That file's contents are printed below.

**********************************************************************************
* If you believe this is due to a bug in scadnano, please file a bug report at   *
*   https://github.com/UC-Davis-molecular-computing/scadnano/issues              *
* Include this entire message in the email.                                      *
**********************************************************************************

stack trace:
TypeError: Cannot read properties of null (reading 'get$iterator')
    at Object.get$iterator$ax (https://scadnano.org/main.dart.js:4085:43)
    at Object.Design__helices_from_json (https://scadnano.org/main.dart.js:40511:19)
    at Object.Design__helices_and_groups_from_json (https://scadnano.org/main.dart.js:40810:15)
    at Object.Design_from_json (https://scadnano.org/main.dart.js:40888:13)
    at Object.load_dna_file_reducer (https://scadnano.org/main.dart.js:25389:48)
    at StaticClosure.app_state_reducer (https://scadnano.org/main.dart.js:33319:18)
    at Store__createReduceAndNotify_closure.call$1 (https://scadnano.org/main.dart.js:70986:28)
    at StaticClosure.oxdna_export_middleware (https://scadnano.org/main.dart.js:40071:12)
    at Store__createDispatchers_closure.call$1 (https://scadnano.org/main.dart.js:70994:34)
    at StaticClosure.zoom_speed_middleware (https://scadnano.org/main.dart.js:23477:12)

pranav-gupta1:


abhishekgokhale:
@pranav-gupta1 Can we also not allow importing .json files when selecting the 'Open' option since open should ideally only allow files with .sc extension

pranav-gupta1:
> 

@abhishekgokhale did you just want me to let the user know in the "error" message (basically just as plain text) or did you want me to change the code such that it is not possible to use a .json file when the "open" option is used?
the first part is essentially done btw.

lib/src/reducers/load_dna_file_reducer.dart
```diff
@@ -43,6 +43,8 @@ ${util.stack_trace_message_bug_report(stack_trace)}''';
         '\n* error type:    ${error.runtimeType}'
         '\n* error message: ${error.toString()}'
         '\n$hline'
+        '\nIf the file is imported from cadnano, use the opton of importing cadnano v2.'
```
dave-doty, line 46: Be more specific than this. Imagine you are the user... they don't know the menu options, so tell them exactly how to import, e.g., "Select File --> Import cadnano v2".

lib/src/reducers/load_dna_file_reducer.dart
```diff
@@ -43,6 +43,8 @@ ${util.stack_trace_message_bug_report(stack_trace)}''';
...
+        '\nImporting a .json file is not allowed when selecting the "Open" option. Use a file with the .sc extension.'
```
dave-doty, line 47: This is a bit confusing because the same error message happens when dragging a `.json` file onto the browser. So they may not have selected File-->Open to see this error message.

dave-doty (changes requested):
This isn't quite what I had in mind. The error message is still confusing because it starts like this:

image

Instead, catch any error that happens while processing the file and then just give a popup (window.alert) explaining that only `.sc` files are accepted.

Here's a good error message to use in this case:

> I was unable to process that file. Only scadnano .sc files are supported for opening via the menu File-->Open or dragging onto the browser. If you are trying to import a cadnano file (ending. in .json), use the menu option File-->Import cadnano v2. Here is the full error message:

Then include the full error message and stack trace below that.

pranav-gupta1:
> This isn't quite what I had in mind. The error message is still confusing because it starts like this:
> 
> image
> 
> Instead, catch any error that happens while processing the file and then just give a popup (window.alert) explaining that only `.sc` files are accepted.
> 
> Here's a good error message to use in this case:
> 
> > I was unable to process that file. Only scadnano .sc files are supported for opening via the menu File-->Open or dragging onto the browser. If you are trying to import a cadnano file (ending. in .json), use the menu option File-->Import cadnano v2. Here is the full error message:
> 
> Then include the full error message and stack trace below that.

Oh okay, understood. I'll work on this.

lib/src/view/menu.dart
```diff
@@ -233,7 +233,7 @@ really want to exit without saving.'''
       DropdownDivider({'key': 'divider-save'}),
       (MenuFormFile()
         ..id = 'import-cadnano-form-file'
-        ..accept = '.json'
+        ..accept = '.sc'
```
dave-doty, line 236: This line is incorrect. The `accept` field should be `'.json'` since that's the file extension for cadnano files.

dave-doty (changes requested):

lib/src/view/menu.dart
pranav-gupta1: Yep, I recently made that change but was not able to push it due to failing the "verify formatting" part of unit test checks. Once I get that working, I will push all the updated code with this change and with the other stuff (pop-up, etc.). Thanks for letting me know.

pranav-gupta1: @dave-doty I pushed the updated changes. Please let me know if I still need to adjust anything.

dave-doty (approved):

dave-doty:
It looks like we never merged this. I tested in the dev branch and this bug is still there.

I think we should merge it instead of closing it, unless I'm misunderstanding something.

dave-doty (approved):

Issue #884, opened 2023-07-21
disallow assigning DNA complements when there are unpaired deletions/insertions


dave-doty:
If there is an unpaired deletion:

![image](https://github.com/UC-Davis-molecular-computing/scadnano/assets/19274365/547fe389-4219-4a5a-abc7-0cfcccc5fc0e)

Then it causes an uninformative error message when trying to assign DNA to one strand and the complement to the other. In this case, clicking on the bottom (purple) strand and assigning AAAAAAAA to it:

![image](https://github.com/UC-Davis-molecular-computing/scadnano/assets/19274365/5603e083-9298-4ea4-9201-1bb572d91a0a)

Results in this error message due to the fact that the other (red) domain has length 7 instead of 8:

![image](https://github.com/UC-Davis-molecular-computing/scadnano/assets/19274365/c6700102-199f-411e-a0bf-f558e2981d6a)

Instead, if there is an unpaired deletion/insertion in any domain that is involved in the DNA sequence assignment, that should be reported as the error, pointing out the exact Address (helix idx, offset) of the unpaired deletion/insertion.

PR #885, opened 2023-07-21
884 disallow assigning dna complements when there are unpaired deletionsinsertions

dave-doty:
## Description
Prevents assigning DNA complements when there is an unpaired insertion or deletion.

## Related Issue
#884 

## Motivation and Context
The error given before this fix was confusing. This more directly indicates the problem.


dave-doty:
Accidently merging to main instead of dev, closing

PR #893, opened 2023-08-04
closes #887 save svg file with image of selected objects

### LINKED ISSUE #887: save SVG file with image of selected objects
This is related to #740. That feature seems very difficult to implement due to browser security restrictions about what types of data can be put on the system clipboard.

This issue is a workaround that is slightly more work to use, but avoids the need to work with the system clipboard. Here, we want to save an SVG file. This is similar to what happens when selecting `Export-->SVG main view`, but the difference is that this will make an SVG file out of only the selected objects (the same as what is currently implemented in PR #769), not the entire main view.

This should be straightforward to implement by combining the logic of collecting the selected object SVG, along with the existing logic from saving SVG files.

This should be a new menu item `Export-->SVG of selected objects`.

abhishekgokhale:


dave-doty (changes requested):
1. (easy) If nothing is selected, pop up a window alert telling them nothing is selected so there is nothing to export, and then don't save any file.
2. (hard) If option View-->Base pairs-->Base pair lines is selected, then the base pair lines of any pair of exported strands should be exported to the SVG as well.

The latter could be difficult because base pair lines are not associated to specific strands, but rather are calculated "globally" in the whole design. Nevertheless, this is something people might want drawn in a figure, so we should figure out how to do this. We can meet to discuss if needed.

3. I noticed that it doesn't work to export selected domains. This is fine for now, mainly I want to be able to select strands and export them. But if it is only supported to do this when strands are selected (not domains, crossovers, ends, etc), then re-title the menu option to "Export SVG of selected strands" to emphasize that this only works when strands are selected.

abhishekgokhale:
I made the first two changes and will have a look for the base pairs one. We can discuss when we meet on Monday if I haven't been able to figure it out by then.

dave-doty:
> I made the first two changes and will have a look for the base pairs one. We can discuss when we meet on Monday if I haven't been able to figure it out by then.

Can you commit the changes? We can make the base pair lines a separate issue to handle in another branch.

abhishekgokhale:
> > I made the first two changes and will have a look for the base pairs one. We can discuss when we meet on Monday if I haven't been able to figure it out by then.
> 
> Can you commit the changes? We can make the base pair lines a separate issue to handle in another branch.

Yes, committed just now

dave-doty (approved):

PR #905, opened 2023-08-19
fixes #902: make menu dropdown wider for Edit-->Copy/Paste/Select

### LINKED ISSUE #902: make menu dropdown wider for Edit-->Copy/Paste/Select
Most of the dropdown menus are properly sized, but Edit-->Copy/Paste/Select is too narrow:

image

It should be wide enough to fit the last two options on one line each.

Note that we are using a library called [React bootstrap](https://react-bootstrap.netlify.app/) for the menu.

EdwinChang24:
The issue was that the `float: right` CSS property on the keyboard shortcut element was causing the element's width to not be considered when computing the entire menu's width. This PR changes dropdown items to use flexbox with `justify-content: space-between` to align the keyboard shortcut to the right.

### DIFF
--- a/lib/src/view/menu_dropdown_item.dart
+++ b/lib/src/view/menu_dropdown_item.dart
@@ -30,10 +30,8 @@ class MenuDropdownItemComponent extends UiComponent2<MenuDropdownItemProps> {
         'disabled': props.disabled,
         'onClick': props.on_click,
       },
-      props.display,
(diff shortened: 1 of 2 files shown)


dave-doty (approved):
Very nice, I would have never figured this out. :)

dave-doty:
This PR looks good to merge whenever you're ready.

PR #954, opened 2023-10-24
953 remove strand in batch

RayBipse:
Fixes batch delete/set labels and names for domains and strands.

## Description
- Used ``BatchAction`` to group together set labels/name.
- Created ``get_selected_domains`` function that returns a ``BuiltSet<Domain>`` set, so setting domain names still work if 
   user selects with domain tool
- Renamed some functions like ``set_substrand_names `` to ``set_domain_names``.

### DIFF
--- a/lib/src/view/design_main_strand.dart
+++ b/lib/src/view/design_main_strand.dart
@@ -481,7 +481,7 @@ assigned, assign the complementary DNA sequence to this strand.
                       (strand) => actions.StrandNameSet(name: null, strand: strand),
                       props.strand,
                       app.state.ui_state.selectables_store.selected_strands,
-                      'remove strand label'))),
+                      'remove strand name'))),
             ContextMenuItem(
               title: 'set domain name',
               on_click: () => set_domain_names(get_selected_domains()),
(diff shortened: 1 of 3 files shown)


dave-doty (changes requested):
This is a good start, but it should also work for extensions and loopouts. For example, if I select these two extensions and select "Remove extension name":

image

it should remove both names, but still removes only one.

If you notice, the original function names talked about substrands rather than domains. I named them that way to make the code generally useful for all three types of substrands (Domain, Extension, and Loopout). Although maybe they weren't being used in the latter two cases, but nevertheless there should be some way (without too much copy-pasting of code) to make this work for all three cases.

lib/src/view/design_main_strand.dart
```diff
@@ -474,10 +477,14 @@ assigned, assign the complementary DNA sequence to this strand.
             if (props.strand.name != null)
               ContextMenuItem(
                   title: 'remove strand name',
-                  on_click: () => app.dispatch(actions.StrandNameSet(name: null, strand: props.strand))),
+                  on_click: () => app.dispatch(batch_if_multiple_selected(
+                      (strand) => actions.StrandNameSet(name: null, strand: strand),
+                      props.strand,
+                      app.state.ui_state.selectables_store.selected_strands,
+                      'remove strand label'))),
```
dave-doty, line 484: I think this should say "remove strand name" instead of "remove strand label"

dave-doty (changes requested):

dave-doty (approved):

PR #957, opened 2023-11-20
 closes #870: Feature request: selection preserves color

### LINKED ISSUE #870: Feature request: selection preserves color
Currently, selecting a strand turns it pink, so the color of that strand can't be known without deselecting it or trying to change it. It would be better for selection to be displayed as a highlight or a thick black outline.

EdwinChang24:
Adds menu option "Retain strand color on selection" in the View menu. When this is enabled, strands are not shown in hot pink when selected, but still have the shadow effect and a slightly larger stroke width. Works by using a different CSS class depending on the option.


dave-doty (changes requested):
Can you make the Dart formatting action pass?

dave-doty (approved):

Issue #897, opened 2023-08-08
option to display paired bases with gray rectangle


dave-doty:
Currently "base pair lines" can be displayed, which helps when making figures:

![image](https://github.com/UC-Davis-molecular-computing/scadnano/assets/19274365/71d8777e-1223-4dc3-840e-adeddc33dd6a)

Another common convention is to depicted double-stranded regions with a gray rectangle connecting the two DNA strands (e.g., from https://www.solo-group.link/papers/dsd-leakless.pdf, Leakless DNA strand displacement systems, Chris Thachuk, Erik Winfree, David Soloveichik, [DNA Computing and Molecular Programming 21 (DNA21), 2015]):

![image](https://github.com/UC-Davis-molecular-computing/scadnano/assets/19274365/cb059658-75f7-4f79-ad15-fdd755a55f8b)

Give this as another view option under View-->Base pairs-->then have a radio button selector (DialogRadio) with three options:

- None
- Lines
- Rectangle

With "... even if bases mismatch" disabled if None is selected.

RayBipse:
Before I implement it, I wanted to double check that I have the correct idea in mind. So we just replace the baselines with a grey rectangle?
<img width="358" alt="Screenshot 2023-12-18 at 12 22 18 PM" src="https://github.com/UC-Davis-molecular-computing/scadnano/assets/46636772/0039586b-b0ef-4a11-b5a3-02b252082ca6">

dave-doty:
[...]

PR #966, opened 2024-03-04
#897 option to display rectangle

RayBipse:
## Description
Created a new display option for base pairs: rectangle.
Moved changed the display into a dialog box.

The dialogue box.
Screenshot 2024-03-03 at 6 43 32 PM
The display when display with rectangle and even if mismatched as false.
Screenshot 2024-03-03 at 6 20 44 PM

### DIFF
--- a/lib/src/actions/actions.dart
+++ b/lib/src/actions/actions.dart
@@ -4012,6 +4012,22 @@ abstract class ExampleDesignsLoad
 }
 
 /////////////////////////////////////////////////////////////////////////////////////////////////////////////
+// pair lines display
+
+abstract class BasePairTypeSet
+    with BuiltJsonSerializable
+    implements Action, Built<BasePairTypeSet, BasePairTypeSetBuilder> {
+  int get selected_idx;
+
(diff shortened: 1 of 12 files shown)


dave-doty (changes requested):
I'd actually prefer a slightly different way of specifying view options.

Instead of popping up a dialog, can we make it all configurable from the submenu View-->Base pairs, similar to View-->Modifications? There, if "Show modifications" is unchecked, that's all you see:

image

but then if you check it, you see the additional options:

image

That's what I had in mind for these options, if unchecked, it looks like it did before, and if checked, then you see something like this directly in the submenu:

image

I don't know if the React Bootstrap library we are using for submenus allows radio buttons, but if not we can just have "lines" and "rectangle" be two checkboxes, and set up the React render code so that exactly one of them is checked (according to the value `AppState.AppUIState.base_pair_display_type`)

RayBipse:
I couldn't find the radio buttons so I just made it into checkboxes like you suggested.

dave-doty (approved):

Issue #1001, opened 2024-10-09
grid side menu is gone


dave-doty:
In the dev branch (not main) the side view menu "grid" is gone"

![image](https://github.com/user-attachments/assets/73c13fb3-2e01-4821-82f5-de1709afd4f7)

How it should be:

![image](https://github.com/user-attachments/assets/b51a9b31-a497-4770-b2a4-61e338f193f3)

dave-doty:
Sorry, I forgot I added Grid as a submenu of Group, so nevermind this issue.

PR #1002, opened 2024-10-09
(Fixes #987) Expand edges of base pair rectangles

### LINKED ISSUE #987: show base pair rectangle all the way to edge of base
The rectangle should cover the whole base, but it only covers half of the bases on the ends of each side:

image

The parts circled in blue should also be gray.

beanbeanjuice:
There was an issue where the left-most and right-most base-pair rectangles would not expand all the way to the edge. This has now been fixed.

## Description
The left-most and right-most squares have been shifted and expanded into rectangles. This fixes issue #987 where there would be some whitespace where the squares don't expand all the way to the edge.

## Related Issue
Fixes #987
To reproduce the issue, click on `View` -> `Base Pairs` then turn **on** `Display as Rectangles`. On the previous commits, this would end up with extra white space being shown at the beginning and end of the base pairs.

## Motivation and Context
This is a bug fix for issue #987.

## How Has This Been Tested?
This has been tested (and works) with each of the example DNA designs. (`File` -> `Load Example`).

Each test has been done on the Arc browser (Chromium based) as well as on Google Chrome.

### Left-Most Base Pair
image

### Right-Most Base Pair
image

### DIFF
--- a/lib/src/view/design_main_base_pair_rectangle.dart
+++ b/lib/src/view/design_main_base_pair_rectangle.dart
@@ -54,6 +54,10 @@ class DesignMainBasePairRectangleComponent extends UiComponent2<DesignMainBasePa
         List<ReactElement> helix_components = [];
         int last_offset = -2;
         var last_svg_forward_pos = null;
+
+        int starting_rect_offset = base_pairs[helix_idx]![1];
+        int ending_rect_offset = base_pairs[helix_idx]!.last;
+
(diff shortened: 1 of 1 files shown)


beanbeanjuice:
This might require a fix for #1001 first? The grid dropdown is currently missing, so it might not work for all cases.

dave-doty:
Can this be taken out of draft and merged?

PR #1007, opened 2024-10-24
Fixed #999 - Right Justified Right Fields in Dialogs with Inputs

### LINKED ISSUE #999: right-justify fields in dialogs
The dialogs are a list of (label, entry field) pairs, e.g., if you set geometric parameters, you see this:

image

It would look nicer for the fields to be right-justified, while the labels remain left-justified.

beanbeanjuice:
## Description
I added CSS to right justify the input in labels for the dialog box.

## Related Issue
Fixes #999. Input is now right-justified.

## Motivation and Context
Fixes #999.

## How Has This Been Tested?
Originally, the fix affected ALL dialogs, but I changed it to only affect dialogs and labels with inputs. This was tested on Windows 11, Chrome.

I am currently unsure of other forms of dialog boxes, but the two that I tested were;
**File -> Load Example**
**Edit -> Set Geometric Parameters**

image

### DIFF
--- a/lib/src/view/design_dialog_form.dart
+++ b/lib/src/view/design_dialog_form.dart
@@ -283,6 +283,7 @@ class DesignDialogFormComponent extends UiStatefulComponent2<DesignDialogFormPro
           ..pattern = r'[+-]?(\d*[.])?\d+' // allow to type floating numbers
           ..value = item.value
           ..step = 'any'
+          ..className = 'dialog-form-item-input'
           ..onChange = (SyntheticFormEvent e) {
             var new_responses = current_responses.toBuilder();
             double? new_value = double.tryParse(e.target.value);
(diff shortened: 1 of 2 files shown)


dave-doty (changes requested):
Make special css class `dialog-item-right-justified-input` for DialogText, DialogFloat, and DialogInteger, add to the list of classes only for those, and CSS format only that new CSS class.

beanbeanjuice:
Please take a look at the new changes. I've added specific CSS styles to only affect certain items. Please let me know if you want me to go the other route of adding a specific class for this instead.

### Before
image
image
image
image


### After
image
image
image
image

dave-doty (approved):
I'm assuming this is what you showed me in the meeting, so I didn't check it myself. But your screenshots look good.

PR #1010, opened 2024-12-05
Fixes #990 - Upgrade to Dart 3

### LINKED ISSUE #990: upgrade to Dart 3
Now that #419 is done (migrating to null safety mode using the latest Dart 2 version 2.19.6), that is the major difficulty in upgrading to Dart 3. But there still may be some issues since it's not backwards compatible. Follow the migration guide here:

https://dart.dev/resources/dart-3-migration

This issue is just just about the bare minimum to make scadnano work under Dart 3. Taking advantage of new features, e.g., [records](#984) will be separate issues.

beanbeanjuice:
## Description
This change adds support for Dart 3. However, there is some dependency overrides added to fix some issues. I have not yet found a way to solve this issue without adding the dependency override.

**Furthermore**, it is still failing `export_cadnano_v2_test.dart`.

## Related Issue
Fixes #990 

## Motivation and Context
This is to hopefully solve the long build times, but most importantly place the repository on the current dart dependency. This brings everything up to date.

## How Has This Been Tested?
This was tested on Windows 11, using **Dart 3.5.4**. The browsers I used were Firefox and Google Chrome. Since the JSONEncoder became a final class, we couldn't extend it anymore. We had to re-implement the JsonEncoder in our own way.

### DIFF
--- a/.github/workflows/dart.yml
+++ b/.github/workflows/dart.yml
@@ -20,7 +20,7 @@ jobs:
       - name: Setup Dart SDK Step 4
         run: sudo sh -c 'wget -qO- https://storage.googleapis.com/download.dartlang.org/linux/debian/dart_stable.list > /etc/apt/sources.list.d/dart_stable.list'
       - name: Setup Dart SDK Step 5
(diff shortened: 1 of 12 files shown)


beanbeanjuice:
It should now be passing all tests. Please let me know if there are any other changes you require.

dave-doty (approved):

PR #1018, opened 2025-03-06
Fixes #984 - Replace Tuples with Records

### LINKED ISSUE #984: replace Tuples with records
Dart 3 has a new feature called records that is similar to tuples/named tuples: https://dart.dev/language/records

We should replace all usages of `Tuple2`, `Tuple3`, etc. from the tuple package with these, and remove the tuple package dependency.

beanbeanjuice:
## Description
This replaces all instances of `TupleX` with Dart's records.

## Related Issue
This fixes #984.

## Motivation and Context
This is much cleaner than using the `Tuple` library, and it was requested by #984.

## How Has This Been Tested?
This was tested using the already existing unit tests. Although, some of the unit tests were modified to accommodate this change.

N/A

### DIFF
--- a/lib/src/middleware/helices_positions_set_based_on_crossovers.dart
+++ b/lib/src/middleware/helices_positions_set_based_on_crossovers.dart
@@ -214,7 +214,7 @@ Map<(int, int), List<(Address, Address)>> _get_addresses_of_selected_crossovers_
     var next_dom = strand.substrands[crossover.next_domain_idx] as Domain;
     var prev_idx = prev_dom.helix;
     var next_idx = next_dom.helix;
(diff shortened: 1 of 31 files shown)


beanbeanjuice (commented):
It looks like all instances of `itemX` were successfully replaced with the corresponding `$X`. It might be best to double check with other eyes just in case.

dave-doty (approved):

beanbeanjuice:
I'll go through and fix all of the merge conflicts.

I will also change it to some more concise code where we can deconstruct the tuple instead. I believe this is documented [here](https://dart.dev/language/records#record-equality).

beanbeanjuice:
Superseded by #1027.

Issue #1019, opened 2025-03-07
upgrade to Dart 3.7 and run dart format


dave-doty:
This is so that we get the new Dart formatting described here: https://github.com/dart-lang/sdk/blob/main/CHANGELOG.md#dart-format

This branch should update the lower bound on dart-sdk to >= 3.7, and also run `dart format -l 110` (double-check the CONTRIBUTING document, but I think that's the correct command) on all files (I think running this from the lib/ directory should run it on all files in all subdirectories.)

PR #1020, opened 2025-03-10
#1019 upgrade to dart 3.7 and run dart format

mongbean:
Dart 3.7 worked well, except one file. "redraw_counter_component_mixin.dart" had an error in line 22 because of the Dart's new update where "_" isn't a variable anymore, and so it can't be used in the next line. I fixed it by changing the variable names (see the commit "dart 3.7, redraw_counter_component_mixin had an error". 
All the formatting stuff is in the format commit.


dave-doty:
Looks like we need to upgrade the Dart version in the Github actions to 3.7 also to make the tests pass.

I'm a bit confused about the `_` issue, I thought what they did is allow you to name multiple variables `_` without clashes, so that you would not need to do what I did in this file and name variables `_`, then another `__`, then another `___`, so that they won't clash. But I'm surprised that it is objecting about simply using `_` at all... what's the issue there?

Perhaps it would be clear if I installed Dart 3.7 and tried for myself but I haven't done that yet. We can talk about this at the next meeting.

dave-doty (approved):

PR #1024, opened 2025-05-02
Fixes #1023 - Update GitHub Actions Dart Version

### LINKED ISSUE #1023: Update GitHub CI to Dart v3.7.3
The tests are currently failing because there is a mismatch between dart versions in the `pubspec.yaml` and the GitHub actions file.

> Note: This may require an update to any README instance of an old dart version, such as `CONTRIBUTING.md`.

beanbeanjuice:
## Description
This updates the dart version in the GitHub actions file(s) to the version specified by `pubspec.yaml`.

## Related Issue
This fixes #1023.

## Motivation and Context
The actions were failing in the dev branch because of a mismatch between the Dart versions and the `pubspec.yaml` versions. The issue is specified at #1023.

## How Has This Been Tested?
I was researching possible solutions to make updating to the latest dart version easier, and came across [this](https://github.com/dart-lang/setup-dart). It's a GitHub action created by the Dart team to automatically download and install the latest (or a specific) version of Dart.

As for testing, you can view the actions below to see that they both work and pass. **However** there is a big caveat. I have also updated the `.github/workflows/gh-pages.yml` and `.github/workflows/gh-pages-dev.yml` with this. I am *unsure* if those work as intended. Further testing **will need** to be done before merging as I do not want to mess up the actual scadnano website.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -241,23 +241,23 @@ git checkout dev
 
 ### Installing Dart
 
-This project requires using Dart version **3.5.4**, not the latest version. Click on a dropdown below for installation instructions for your operating system.
+This project requires using Dart version **3.7.3**, not the latest version. Click on a dropdown below for installation instructions for your operating system.
 
 <!--TODO: Find a way to use code blocks with syntax highlighting inside <details>-->
 
 <details><summary><strong>Windows</strong></summary>
 First, install <a href="https://chocolatey.org/install">Chocolatey</a> if you haven't already. If <code>choco help</code> shows a help menu for using Chocolatey, then you've set it up correctly.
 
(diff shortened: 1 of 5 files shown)


CONTRIBUTING.md
```diff
@@ -241,23 +241,23 @@ git checkout dev
 
 ### Installing Dart
 
-This project requires using Dart version **3.5.4**, not the latest version. Click on a dropdown below for installation instructions for your operating system.
+This project requires using Dart version **3.7.3**, not the latest version. Click on a dropdown below for installation instructions for your operating system.
```
dave-doty, line 244: Let's change this. These instructions were for when we did not have null safety. Now I'd rather if the instructions are always to use the latest Dart. If we run into another problem where we don't seem to be able to upgrade, I'll revise these instructions. But for now please write it to indicate how to get the latest version of dart. But leave these instructions as a separate subsection later titled "Installing a particular version of dart".

dave-doty (changes requested):

dave-doty (approved):

beanbeanjuice:
I also updated the GH actions file(s) to always use the *latest* dart version as well.

PR #1103, opened 2026-08-02
1102 replace automated release action

### LINKED ISSUE #1102: replace automated release action
Do everything that was done here: https://github.com/UC-Davis-molecular-computing/scadnano-python-package/issues/323

That issue description is incomplete; look to the actual actions and github repo configuration (e.g., in the end we decided to keep main as the default branch), and mimic what was done for that issue. Unfortunately it was not done on a separate branch but hopefully it is clear just from reading the github actions and CONTRIBUTING document how to mimic that in this repo.

dave-doty:
Fixes #1102


dave-doty:
Thanks @copilot-pull-request-reviewer — 3 of the 5 inline findings were valid and are fixed in b21645a4; the other 2, plus one of the 3 suppressed comments in the second pass, are incorrect. Recording the evidence here so they don't get re-raised.

## Fixed

- **`branch-for-issue.sh:34`** — correct, `^[0-9]+$` accepted `0` and `007` while the error message promised a positive integer. Now `^[1-9][0-9]*$`. (The `.ps1` already got this right via `[ValidateRange(1, [int]::MaxValue)]`.)
- **`branch-for-issue.ps1:60`** — good catch, and a real bug. It tested `$PSScriptRoot\.git`, but `gh issue develop` infers the repository from the **current** directory, so running the script from elsewhere would pass the check and then fail confusingly. Now uses `git rev-parse --git-dir`, matching the `.sh`. Verified: passes inside the working copy, errors outside it.
- **`CONTRIBUTING.md:786`** — "Github" → "GitHub".

## Not applying

**`concurrency` does not support a `queue` key.** It does. From the [workflow syntax reference](https://docs.github.com/en/actions/reference/workflows-and-actions/workflow-syntax#concurrency):

> To allow more than one `pending` job or workflow run to wait in the same concurrency group, use the optional `queue` property. […] `max`: Up to 100 jobs or workflow runs can be `pending` in the concurrency group. […] The combination of `queue: max` and `cancel-in-progress: true` is not allowed and will result in a workflow validation error.

This workflow uses `queue: max` with `cancel-in-progress: false`, which is the supported combination. The default (`queue: single`) allows only **one** pending run and silently cancels it when a third arrives — for the sole creator of releases that would mean a lost release with no red X, which is exactly what `queue: max` prevents.

**`retarget-pr-to-dev.yml` needs `issues: write`.** It does not. Checked empirically rather than by reading docs: the identical workflow in the sibling repo `scadnano-python-package` runs with only `pull-requests: write`, and run [30712662045](https://github.com/UC-Davis-molecular-computing/scadnano-python-package/actions/runs/30712662045) concluded **success** and posted the comment now visible on that repo's PR #342 (`github-actions[bot]`, 18:30:19Z). `pull-requests: write` covers `issues.createComment` when the target number is a pull request.

**`dependabot.yml` should use `directory:` not `directories:`.** `directories` is supported. From the [Dependabot options reference](https://docs.github.com/en/code-security/dependabot/working-with-dependabot/dependabot-options-reference):

> Use `directories` to define a list of multiple directories of manifests. […] The `directories` key supports globbing and the wildcard character `*`. These features are not supported by the `directory` key.

It is needed here specifically because Dependabot's `github-actions` ecosystem at `/` scans only `.github/workflows` plus a root-level `action.yml` — it does not recurse into `.github/actions/`, so the local composite action `.github/actions/setup-dart` has to be listed explicitly or `dart-lang/setup-dart` would never be updated.

All three of these are newer platform features, which suggests a knowledge-cutoff issue rather than a reasoning one.

